PR #7, opened 2018-10-15
Add how to build the application to the README

AlexanderThaller:
Just a small section on how to actually build `ttyplot` if neccessary.

The installation information for `ncurses` I took from https://www.ostechnix.com/how-to-install-ncurses-library-in-linux/.

I hope the installation instructions for the other operating systems (that are not ArchLinux) are correct.

I tested it under ArchLinux and there it was fine.

Hope this helps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+/ttyplot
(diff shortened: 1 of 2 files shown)


tenox7:
People who do not know how to install and compile with curses should download and use a pre-compiled binary which I provide. Also I have added a Makefile so build instructions are not required any more.

PR #11, opened 2018-10-15
add simple Makefile

netzbasis:
Add a Makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,13 @@
+NAME = ttyplot
+CC = cc
+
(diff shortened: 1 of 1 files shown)


tenox7:
I had one before just forgot to add it

netzbasis:
It's simple enough. Also ttyplot already made it into OpenBSD ports and packages... so better do a 1.1 release with the fixes so far, instead of changing tags or updating release tarballs.

misuchiru03:
Yes. Thank you for the clarification.

PR #14, opened 2018-10-18
sensors temp, gitignore

spacerace:
This for now, don't have much spare time right now... Will look at this scanf-thing I mentioned soon...


tenox7:
please remove the lm-sensors temperature and just leave gitignore for now

PR #16, opened 2018-10-21
Makefile: Dehardcode compiler and cflags

Johnnynator:
It is generally nicer to not hard code the compiler and cflags so that the Makefile works in more use cases e.g. cross compiling, systems that have clang installed and so on.
Note `$(CC)` defaults to gcc if it is unset.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,10 @@
 
+CFLAGS = -Wall -O3
+
 all: ttyplot
 
 ttyplot: ttyplot.c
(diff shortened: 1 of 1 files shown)


tenox7:
Shouldn't -lcurses go to LDFLAGS?

Johnnynator:
@tenox7 LDFLAGS and CFLAGS should be used in a way that they can be overwritten from the outside without being able to break, so they should only contain optional flags. -lcurses is mandatory, sometimes libraries do get put into a variable called LIBS.

PR #21, opened 2019-02-16
Add clipping char option (-C)

gromgit:
This switches to a different character to highlight values that have
exceeded a user-defined max value. The original implementation just
leaves a blank line instead, which is ambiguous but may be desired
in various applications.

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,13 +134,15 @@ options
 =======
 
 ```
-ttyplot [-r] [-c plotchar] [-s softmax] [-m hardmax] [-t title] [-u unit]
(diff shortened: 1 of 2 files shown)


tenox7:
Excellent. I would make it a default as the current behavior is pretty lame. Could you make it default to 'x' for example?

gromgit:
> Excellent. I would make it a default as the current behavior is pretty lame. Could you make it default to 'x' for example?

Done. I've updated this PR accordingly.

tenox7:
Awesome thank you!

PR #33, opened 2019-05-02
Makefile tweaks

sjmulder:
- Respect system CFLAGS by adding to it, rather than overriding.

 - Use built-in inference rules for compilation. It's shorter but technically the defaults may be better suited to the system or build environment.

- Don't override default optimization level. Except in specific cases the system wide configuration should be respected.

 - Mark 'clean' as a phony target, which is to say that it's not a file target so the existence of a 'clean' file doesn't satisfy the target.

All matters of taste though so no hurt feelings if you disagree!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
-
-CFLAGS = -Wall -O3
+CFLAGS += -Wall -O3
 
 all: ttyplot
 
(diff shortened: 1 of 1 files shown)


tenox7:
I like it as it's much shorter. Thank you!

PR #35, opened 2019-05-02
Add install and uninstall targets

sjmulder:
PREFIX is where the program is found after installation, e.g. /usr or /usr/local.

DESTDIR can be used by package build tools and such as a staging directory from which the package is created.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,5 @@
+DESTDIR ?=
+PREFIX  ?= /usr/local
 
 CFLAGS = -Wall -O3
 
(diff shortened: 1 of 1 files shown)


tenox7:
Thank you. Could you also add targets to build rpm and deb if it's not too much to ask?

sjmulder:
I'm afraid that's not as straightforward - in any case I'm not up to the task. You might want to check out [packagecore](https://github.com/BytePackager/packagecore) but ideally the distributions will pick it up and package it.

In fact this PR is prep for packaging in [pkgsrc](http://www.pkgsrc.org/), a cross platform package manager.

tenox7:
Can you update it so there is no conflict with the new Makefile format. My bad I probably should merged this first.

sjmulder:
The PRs conflicted when I made them; a trade off for not wanting one to depend on the other.

I've now also added the man page to the install target.

sjmulder:
Any issues left?

tenox7:
none, thanks!

PR #37, opened 2019-05-07
Add pledge() support

sjmulder:
initscr() is moved up so we don't need rpath for curses accessing its
terminal database.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -14,6 +14,10 @@
 #include <curses.h>
 #include <signal.h>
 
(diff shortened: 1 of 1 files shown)


tenox7:
This doesn't work for displaying usage info, eg. when running `ttyplot -h`. This is because usage() printf()s to curses initialized screen. I'm fine with adding pledge() but perhaps after getopt loop?

sjmulder:
Ouch, good catch. I should've tested a bit more.

PR #43, opened 2019-10-02
keep output visible after exiting with ^c or there is no more data

### LINKED ISSUE #41: don't clear screen when exiting
it would be nice to leave output after exiting with ^c or when there is no more data

this can be done by suing pad or newterm instead of initscr

references:
https://stackoverflow.com/questions/4772061/how-do-i-use-getch-from-curses-without-clearing-the-screen
https://stackoverflow.com/questions/14446311/make-curses-program-output-persist-in-terminal-scrollback-history-after-program

justin2004:
resolves #41

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -34,6 +34,9 @@
 #define T_LLCR ACS_LLCORNER
 #endif
 
+/* global because we need it accessible in the signal handler */
+SCREEN *sp;
+
 void usage() {
     printf("Usage:\n  ttyplot [-2] [-r] [-c plotchar] [-s scale] [-m max] [-t title] [-u unit]\n\n"
             "  -2 read two values and draw two plots, the second one is in reverse video\n"
@@ -123,6 +126,7 @@ void finish() {
     echo();
     refresh();
     endwin();
+    delscreen(sp);
     exit(0);
 }
 
@@ -183,7 +187,7 @@ int main(int argc, char *argv[]) {
                 break;
         }
 
-    initscr(); /* uses filesystem, so before pledge */
+    sp = newterm("linux", stdout, stdin);
 
     #ifdef __OpenBSD__
     if (pledge("stdio tty", NULL) == -1)
@@ -327,5 +331,6 @@ int main(int argc, char *argv[]) {
     }
 
     endwin();
+    delscreen(sp);
     return 0;
 }


justin2004:
works on my linux machines but i tested this on my macbook and it did not work. so i will need to update the term type.

justin2004:
just to gather some information i wrote a script to see which term types work (leave the output visible).

maybe we include a list of working term types and only call newterm() when an environment already has TERM set to one of the working term types?   otherwise we just use initscr().

```
n cygwin
n cons25
n cons25-debian
n screen-256color
n screen-256color-bce
n screen-w
n screen-bce
n screen
n screen-s
y sun
y linux
y Eterm
y xterm-vt220
y xterm-r5
y xterm-r6
y xterm-color
y xterm-256color
y xterm-mono
y xterm
y xterm-xfree86
n mach-gnu
n mach-gnu-color
y mach-bold
y mach-color
y mach
y vt100
y vt102
y vt220
n vt52
n dumb
n ansi
y wsvt25
y wsvt25m
y rxvt
y rxvt-basic
y rxvt-unicode
n pcansi
n hurd
n cons30-m
n cons30
y cs10
n citoh-6lpi
n cdc456
n cons43
n cit101e-rv
n c108-w
n cons60l1-m
n cdc752
n cons50-m
n cons50r-m
y cit500
n contel300
n cad68-3
y color_xterm
n cygwinDBG
n cons50l1
n cit101e-n132
y cs10-w
n c108-rv-4p
n citoh-prop
n cit101
n cons60r-m
n cit101e-n
n cci
n c108-rv
n cons60l1
broke my script citoh-pica
```

tenox7:
Thank you. I would definitely want to stay away from testing terminal types in the app. This is the role of ncurses / any library to do it for you. I think perhaps some specific capabilities need to be tested or modes set instead?

ttyplot.c
```diff
@@ -183,7 +187,7 @@ int main(int argc, char *argv[]) {
                 break;
         }
 
-    initscr(); /* uses filesystem, so before pledge */
+    sp = newterm("linux", stdout, stdin);
```
tenox7: This definitely should not be "linux" type. You are hardcoding terminal type. Null perhaps?

ttyplot.c
```diff
@@ -123,6 +126,7 @@ void finish() {
     echo();
     refresh();
     endwin();
+    delscreen(sp);
```
tenox7: What about resize() function?

ttyplot.c
tenox7: I'm also now thinking that with hardcoded terminal type your test results are probably not conclusive.

tenox7:
I may be completely mistaken but upon further investigation I have determined that this problem has absolutely nothing to do with initscr() vs newterm().

Instead the problem is with xterm (and other terminal) "alternate screen" switching.

This can be worked around in code by implementing an ugly hack from dialog(1) that disables processing enter/exit alt screen. I actually have working solution but currently cannot publish due to incompatible licensing.

Another method is simply to use a different, but still compatible terminal definition, eg 

```
seq 10 | TERM=vt100 ttyplot
```

seem to work fine on xterm-ish terminals.

justin2004:
> Another method is simply to use a different, but still compatible terminal definition

that is similar to what i am working on. 

i often used tmux so my TERM is usually "screen."
but it does not work like we want with newterm().

but i see that vt100 has capability smam (automatic margin mode) and screen does not have that capability, for example. 

so if i can find the single or handful of capabilities all the working term types have in common we could just, at runtime, see what TERM is bound to and if the user's current TERM has the desirable capabilities then we use newterm() otherwise we just use initscr() and they don't get this "persist the output" feature.

i am looking into that because i am not sure how reliable it is just to change TERM.

it feels more reliable to just say "oh, your current TERM has the right capability so you get the feature otherwise you just get your screen cleared on exit."

tenox7:
Try this:

```
infocmp -I $TERM | sed -e 's/smcup=[^,]*,//g' -e 's/rmcup=[^,]*,//g' | tic -
```

:)

tenox7:
I have done even more testing and confirmed work around either by specifying different terminal type or patching terminfo works just fine. Changing initscr() to newterm() has nothing to do it. It was probably my poor understanding at that time.

Issue #42, opened 2019-08-26
feature request: set minimum


EpicWink:
Option to set vertical axis minimum.

Currently, the minimum on the vertical axis is 0. If the value plotted fluctuates around a high value, a lot of the resolution of the fluctuations is lost due to most of the graph representing the baseline.

This is a proposal to have a command-line parameter to set the minimum of the vertical scale (similar to the maximum value, I suppose)

EpicWink:
I always believe in keeping the API consistent, so yes, corresponding 'minimum' versions of `-s` (initial maximum / softmax) and `-m` (error-above maximum / hardmax)

bratekarate:
I see that the PR for this issue was closed a while ago. Are there any plans to implement this/use the code from the closed PR?

tenox7:
Yeah I would, unsure why the PR was closed. Perhaps OP could shed some light on this?

EpicWink:
This issue's OP has no coverage on the associated PR :)

tenox7:
the pr is merged!

Issue #22, opened 2019-02-16
Please add options softmin and hardmin


emk2203:
During real-life use, it turns out that having minimum options would be convenient. Could they be added?

For something exceeding the max and min values, I propose to have a full bar with some marker for max and not drawing anything at all for min, to distinguish them.

gromgit:
>For something exceeding the max and min values, I propose to have a full bar with some marker for max and not drawing anything at all for min, to distinguish them.

I submitted PR #21 just before you posted, adding a `-C` (clipping char) option to draw bars with a different character when the plot's max value is exceeded. For example:
```
while true; do echo ${RANDOM} ${RANDOM} | ttyplot -2 -C @ -m 2000 -t "random values"
```

hartwork:
@tenox7 can this be closed as fixed by #47 or is anything still missing?

PR #47, opened 2020-04-23
Implemented hardmin feature

hrshbh:
Implemented the hardmin feature, as requested in issue #42 and issue #22 , along with other minor improvements.
softmin **not** implemented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -171,8 +171,10 @@ flags
   -r rate of a counter (divide value by measured sample interval)
   -c character to use for plot line, eg @ # % . etc
   -e character to use for plot error line when value exceeds hardmax (default: e)
+  -E character to use for error symbol displayed when value is less than hardmin (default: v)
   -s initial scale of the plot (can go above if data input has larger value)
   -m maximum value, if exceeded draws error line (see -e), plot scale is fixed
+  -M minimum value, if entered less than this, draws error symbol (see -E), lower-limit of the plot scale is fixed
   -t title of the plot
   -u unit displayed beside vertical bar
 ```
--- a/README.md
+++ b/README.md
@@ -165,7 +165,7 @@ flags
 =====
 
 ```
-  ttyplot [-2] [-r] [-c plotchar] [-s scale] [-m max] [-t title] [-u unit]
+  ttyplot [-2] [-r] [-c plotchar] [-s scale] [-m max] [-M min] [-t title] [-u unit]
 
   -2 read two values and draw two plots, the second one is in reverse video
   -r rate of a counter (divide value by measured sample interval)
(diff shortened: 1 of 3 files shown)


tenox7:
Very nice! However you have also made some changes that do not concern hardmin feature. This makes it hard to read. Please split this PR in to smaller ones. Thank you.

hrishabh23:
Done. Now all changes strictly concern with hardmin.

tenox7:
I would also like the other changes, just in a separate PR. Thank you!

README.md
```diff
@@ -173,6 +173,7 @@ flags
   -e character to use for plot error line when value exceeds hardmax (default: e)
   -s initial scale of the plot (can go above if data input has larger value)
   -m maximum value, if exceeded draws error line (see -e), plot scale is fixed
+  -M minimum value, if not exceeded draws error line (see -e), plot scale is fixed
```
tenox7, line 176: I think this it little bit unfortunate wording. Doesn't really explain whats going on there. I think it should say something about if value is lower than minimum then.... It should also say that the min or lower scale is fixed, as max is not. This also begs to fix line above to make it more specific about upper/max scale.

tenox7: Also upon testing I'm not convinced that printing error line throughout the whole vertical scale is appropriate in this case. This makes sense for > max but not for < min. I think maybe print 'm' or just "min", or something like "v" or down arrow on the bottom of the scale would be better?

ttyplot.1
```diff
@@ -53,6 +53,14 @@ as a hard value limit after which an error line will be drawn
 see
 .Fl e
 .Pc .
+.It Fl M Ar hardmin
+Use
+.Ar hardmin
+as a hard value limit before which an error line will be drawn
```
tenox7, line 59: this is also rather unfortunate explanation, please word it better and explain what exactly will go on

ttyplot.c
```diff
@@ -172,6 +175,14 @@ int main(int argc, char *argv[]) {
...
                 hardmax=atof(optarg);
                 break;
+            case 'M':
+                hardmin=atof(optarg);
+                int i;
+                for(i=0;i<1024;i++){
+                    values1[i]=hardmin;
+                    values2[i]=hardmin;
+                }
+                break;
```
tenox7, line 185: I think also that if hardmin is set also softmax should be set to hardmin+1 or so. If you run `ttyplot -M 10` and enter few values < 10 in the beginning the plot will behave weirdly because min=max. If you run `ttyplot -M 10 -s 11` it will work much better.

hrishabh23:
I agree, I was sloppy about the `usage()` and the manpage. I've tried to make them better, but if still changes are required in them, then I think you can amend it here before merging.

Regarding the `softmin` matter, I have added checks after parsing of args.

And as to the error response of value being less than `hardmin`, I was thinking of making as little changes as possible to the code, as it was beautiful already. Now, I've added a new error symbol, and included option to modify it.

tenox7:
Hi, why did you close this? I would really like to get min/hardmin in to ttyplot.

hrshbh:
Hi, the PR was in the open state for quite some time, without any response. So, I decided to close it.
Should I reopen this?
It's been almost a year, probably there will be some merge conflicts that will need to be resolved first.

tenox7:
I really want to have it so I guess reopen it?

hrshbh:
Thanks for the interest.
Will soon work to resolve the merge conflicts, if any.

ttyplot.c
hrshbh: Added the check after opt parsing.

ttyplot.1
hrshbh: Improved manual a bit. Inform me if any more changes are needed.

tenox7:
thanks! let me try it!

tenox7 (approved):

tenox7:
Merged thank you very much!

PR #54, opened 2021-05-17
Debianize

Elbandi:
This is the right debian package build files


tenox7:
This is sweet thank you! The question is if you can help to put it in debian repo aka mentors stuff?

tenox7:
this is no longer required as the package for debian has been released and has a maintainer

PR #67, opened 2023-03-12
Makefile: Use pkg-config for systems where ncurses is not libcurses.so (fixes #66)

### LINKED ISSUE #66: [master] Fails to compile with error: cannot find -lcurses: No such file or directory
```console
# make
cc -Wall -Wextra    ttyplot.c  -lcurses -ltinfo -o ttyplot
/usr/lib/gcc/x86_64-pc-linux-gnu/11/../../../../x86_64-pc-linux-gnu/bin/ld: cannot find -lcurses: No such file or directory
collect2: error: ld returned 1 exit status
make: *** [<builtin>: ttyplot] Error 1

# pkg-config --libs ncurses
-lncurses -ltinfo
  ^
```

Pull request upcoming...

hartwork:
Fixes #66

I'm happy to adjust and discuss.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
 CFLAGS += -Wall -Wextra
-LDLIBS += -lcurses -ltinfo
+LDLIBS += `pkg-config --libs ncurses 2>/dev/null || echo '-lcurses -ltinfo'`
 PKG = ttyplot_1.4-1
 PKGDIR = $(PKG)/usr/local/bin
 


tenox7:
Running make on macOS returns:

```
/bin/sh: pkg-config: command not found
```

hartwork:
@tenox7 two thoughts on that:
- they could install it via homebrew
- we could try something like `pkg-config --libs ncurses 2>/dev/null || echo '-lcurses -ltinfo'` instead?

hartwork:
@tenox7 update to demo the fallback approach

tenox7:
that looks better, let me test it a bit :)

hartwork:
@tenox7 any news?

tenox7:
BTW the variable expansion is done when CC command is invoked

```
$ make
cc -O2 -pipe -Wall -Wextra  ttyplot.c `pkg-config --libs ncurses 2>/dev/null || echo '-lcurses -ltinfo'` -o ttyplot
```

tenox7:
I guess it's not a problem on modern systems.

PR #69, opened 2023-03-12
Makefile: Fix broken "make uninstall"

hartwork:
- Fix use of `$(PREFIX)` where `$(DESTDIR)` was intended
- Remove empty folders (and those only)


hartwork:
@tenox7 do you have a minute for review? I can drop the `rmdir`s if it's too much.

tenox7 (commented):
LGTM

tenox7:
Sorry I just noticed the `rmdirs` now. Removed them past the PR.

PR #70, opened 2023-03-12
Fix signal handler function signatures

hartwork:
Previously, e.g. GCC 11 was warning:
> ```console
> # CFLAGS='-std=gnu99 -pedantic' make
> cc -std=gnu99 -pedantic -Wall -Wextra    ttyplot.c  -lncurses -ltinfo -o ttyplot
> ttyplot.c: In function ‘main’:
> ttyplot.c:291:22: warning: ISO C forbids conversion of function pointer to object pointer type [-Wpedantic]
>   291 |     signal(SIGWINCH, (void*)resize);
>       |                      ^
> ttyplot.c:291:22: warning: ISO C forbids passing argument 2 of ‘signal’ between function pointer and ‘void *’ [-Wpedantic]
>   291 |     signal(SIGWINCH, (void*)resize);
>       |                      ^~~~~~~~~~~~~
> In file included from ttyplot.c:15:
> /usr/include/signal.h:88:57: note: expected ‘__sighandler_t’ {aka ‘void (*)(int)’} but argument is of type ‘void *’
>    88 | extern __sighandler_t signal (int __sig, __sighandler_t __handler)
>       |                                          ~~~~~~~~~~~~~~~^~~~~~~~~
> ttyplot.c:292:20: warning: ISO C forbids conversion of function pointer to object pointer type [-Wpedantic]
>   292 |     signal(SIGINT, (void*)finish);
>       |                    ^
> ttyplot.c:292:20: warning: ISO C forbids passing argument 2 of ‘signal’ between function pointer and ‘void *’ [-Wpedantic]
>   292 |     signal(SIGINT, (void*)finish);
>       |                    ^~~~~~~~~~~~~
> In file included from ttyplot.c:15:
> /usr/include/signal.h:88:57: note: expected ‘__sighandler_t’ {aka ‘void (*)(int)’} but argument is of type ‘void *’
>    88 | extern __sighandler_t signal (int __sig, __sighandler_t __handler)
>       |                                          ~~~~~~~~~~~~~~~^~~~~~~~~
> ```

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -199,7 +199,8 @@ void paint_plot() {
     sigprocmask(SIG_UNBLOCK, &sigmsk, NULL);
 }
 
-void resize() {
+void resize(int signum) {
+    (void)signum;
     sigprocmask(SIG_BLOCK, &sigmsk, NULL);
     endwin();
     refresh();
@@ -208,7 +209,8 @@ void resize() {
     paint_plot();
 }
 
-void finish() {
+void finish(int signum) {
+    (void)signum;
     sigprocmask(SIG_BLOCK, &sigmsk, NULL);
     curs_set(FALSE);
     echo();
@@ -288,8 +290,8 @@ int main(int argc, char *argv[]) {
     mvprintw(height/2, (width/2)-14, "waiting for data from stdin");
     refresh();
 
-    signal(SIGWINCH, (void*)resize);
-    signal(SIGINT, (void*)finish);
+    signal(SIGWINCH, resize);
+    signal(SIGINT, finish);
     sigemptyset(&sigmsk);
     sigaddset(&sigmsk, SIGWINCH);
 


tenox7:
thank you for the submission, I need to verify how that works on older unixes

hartwork:
Okay sure. The OpenGroup POSIX page for `signal` is at https://pubs.opengroup.org/onlinepubs/009604499/functions/signal.html . Are we talking POSIX?

hartwork:
PS: My local `man 2 signal` says:
> ```
> STANDARDS
>        POSIX.1-2001, POSIX.1-2008, C99.
> ```
If C99 is reality, that would be great.

ttyplot.c
```diff
@@ -199,7 +199,8 @@ void paint_plot() {
     sigprocmask(SIG_UNBLOCK, &sigmsk, NULL);
 }
 
-void resize() {
+void resize(int signum) {
+    (void)signum;
```
tenox7, line 203: OK I understand that a handler can receive int signal number, but I'm not sure what does signum on line 203 do?

tenox7 (commented):
can you explain what does  (void)signum; do here?

hartwork:
> can you explain what does (void)signum; do here?

@tenox7 that's most portable way to mark a variable as unused I know. There are other ways but probably less portable, e.g. see https://stackoverflow.com/questions/3599160/how-can-i-suppress-unused-parameter-warnings-in-c . What do you think?

ttyplot.c
hartwork: Now answered at https://github.com/tenox7/ttyplot/pull/70#issuecomment-1465485595 (for anyone else coming here)

tenox7:
I see. Thank you! Let me look on how it works on other systems.

PR #71, opened 2023-03-12
Address warning `-Wstrict-prototypes` e.g. for Clang 16

hartwork:
Previously:
> ```console
> # CFLAGS='-std=gnu99 -pedantic' make CC=clang
> clang -std=gnu99 -pedantic -Wall -Wextra    ttyplot.c  `pkg-config --libs ncurses` -o ttyplot
> ttyplot.c:51:11: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
> void usage() {
>           ^
>            void
> ttyplot.c:87:11: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
> void gethw() {
>           ^
>            void
> ttyplot.c:142:16: warning: a function declaration without a prototype is deprecated in all versions of C [-Wstrict-prototypes]
> void paint_plot() {
>                ^
>                 void
> 3 warnings generated.
> ```

PR #72, opened 2023-03-12
Make GitHub Actions build on Linux (to detect and prevent future build regressions)

hartwork:
This sits on top of pull requests #67, #69, #70, #71 to have a chance for a green CI. The idea is to first get all those merged, and then I would rebase this onto latest `master` to cut it down to the essence which is adding CI.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,15 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under Apache License Version 2.0
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    commit-message:
+      include: "scope"
+      prefix: "Actions"
+    directory: "/"
+    labels:
+      - "enhancement"
+    schedule:
+      interval: "weekly"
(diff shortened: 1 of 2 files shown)


hartwork:
Rebased onto latest `master` now. PS: This very Ci can be seen in action at https://github.com/hartwork/ttyplot/actions .

hartwork:
@tenox7 rebased, only sits on top of #67 now.

hartwork:
@tenox7 do you have a minute for review?

Makefile
```diff
@@ -2,7 +2,7 @@ DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
 CFLAGS += -Wall -Wextra
-LDLIBS += -lcurses -ltinfo
+LDLIBS += `pkg-config --libs ncurses 2>/dev/null || echo '-lcurses -ltinfo'`
```
tenox7, line 5: this will not work on a system that doesn't have pkg-config

hartwork: @tenox7 how will it not? That's what the `||` is for.

PS: Having a build require pkg-config would not be something crazy to ask from users also. It's about as standard of a build tool as it gets.

hartwork:
Rebased off of #67 and onto latest `master` now.

hartwork:
@tenox7 we're in three weeks here now without any real progress. If this pull request is not welcome, now would be a good time to say, and we'll just close it and I'll move on. How would you like to proceed?

tenox7:
PRs are most welcome. I'm mostly busy with other projects and I didn't want to risk breaking things. Never used github actions before so not sure what to expect from this. I'll merge it.

Issue #78, opened 2023-03-14
Minimum window size assumptions can be violated?


hartwork:
Hi!

From a user point of view, resizing ttyplot's window to small sizes and back can get ttyplot into a broken state, e.g. see video below.

From a code point of view, there are multiple places where a certain minimum size is assumed…

- `plotheight=height-4;`
- `(width/2)-14`
- `(width/2)-(strlen(title)/2)`

…but not checked for. So the assumptions can be violated in practice, e.g. during a resize.

[python-prompt-toolkit](https://github.com/prompt-toolkit/python-prompt-toolkit) addresses this problem by displaying something like "window too small" whenever that's the case. I think that would be great to have here for robustness, too.

What do you think?

PS: This may or may not be related to #77.

https://user-images.githubusercontent.com/1577132/224890903-a074ab52-20f5-46df-a135-83e2e6e6183d.mp4

tenox7:
Yeah great idea thank you!

hartwork:
This is still broken with latest 1.5.1. I'll adjust the title accordingly…

hartwork:
It turns that `/dev/random` triggers multiple things — including undefined behavior — needs more investigation.
For the simple part — displaying a message when then window is too small — pull request #97 is now ready for review.

hartwork:
Closing as fixed by #97 in release 1.5.2; for the undefined behavior as aspect there is dedicated issue #111 by now…

PR #97, opened 2023-11-12
Display "Window too small..." when the terminal window is too small (addresses part of #78)

hartwork:
Addresses part of #78

Depends on #94 for a fully green CI.

Looking forward to review :pray: 

PS: The current indentation is arguably and meant to be temporary and only to help the first round of review.


tenox7:
LGTM thank you!

PR #99, opened 2023-11-13
Support plotting with multibyte characters

### LINKED ISSUE #84: Multi-byte characters not supported?
I'm using ttyplot and I LOVE it, but whenever I set -c to one of https://en.wikipedia.org/wiki/Block_Elements, the display becomes kaleidoscopic. So I have the suspicion that multi-byte characters are not properly supported?

edgar-bonet:
This pull request addresses issue #84. Note that, in order for ncurses to support non-ASCII characters, we need to:

* Ask for wide character support by defining a suitable macro before including `<curses.h>`. I went for `_XOPEN_SOURCE`, but we could use the more explicit `NCURSES_WIDECHAR` instead.

* Link with the version of the library that provides this support. On my system (Ubuntu 22.04), this is ncursesw.so. Some tests should be done on other OSes in order to find the suitable link options.

One drawback of using the wide character version of `mvvline()` is that it does not support ACS (alternative character set) characters. Since we cannot draw with `ACS_VLINE`, we default to the visually equivalent “│” (U+2502 box drawings light vertical). This, however, cannot be done on the C locale (or any other 8-bit locale). In this case we use the ASCII character “|” (U+007C vertical line), which unfortunately leaves tiny gaps between successive character cells.

With this PR, the command:

```shell
{ ./torture | head -75; sleep 1m; } | ./ttyplot -c ╳
```

displays:

screenshot

I tested with multiple plotting characters, inside and outside the BMP, and also with ASCII characters in the C locale. It works for me, but more testing would be welcome, especially on OSes other than Ubuntu.

Fixes #84.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
 CFLAGS += -Wall -Wextra
+CFLAGS += `pkg-config --cflags ncursesw`
 LDLIBS += `pkg-config --libs ncursesw`
 torture: LDLIBS = -lm
 
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
 CFLAGS += -Wall -Wextra
-LDLIBS += `pkg-config --libs ncurses`
+LDLIBS += `pkg-config --libs ncursesw`
 torture: LDLIBS = -lm
 
 all: ttyplot
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -5,6 +5,11 @@
 // Apache License 2.0
 //
 
+// This is needed on macOS to get the ncurses widechar API, and pkg-config fails to define it.
+#ifdef __APPLE__
+#define _XOPEN_SOURCE_EXTENDED
+#endif
+
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -5,12 +5,14 @@
 // Apache License 2.0
 //
 
+#define _XOPEN_SOURCE 500  // Get ncurses wchar_t support from SUSv2 (UNIX 98)
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
 #include <unistd.h>
 #include <float.h>
 #include <time.h>
+#include <locale.h>
 #include <ncurses.h>
 #include <signal.h>
 #include <errno.h>
@@ -45,7 +47,7 @@
 #endif
 
 sigset_t sigmsk;
-chtype plotchar, max_errchar, min_errchar;
+cchar_t plotchar, max_errchar, min_errchar;
 time_t t1,t2,td;
 struct tm *lt;
 double max=FLT_MIN;
@@ -121,19 +123,26 @@ void draw_axes(int h, int ph, int pw, double max, double min, char *unit) {
     mvaddch(h-3, 2, T_LLCR);
 }
 
-void draw_line(int x, int ph, int l1, int l2, chtype c1, chtype c2, chtype hce, chtype lce) {
+void draw_line(int x, int ph, int l1, int l2, cchar_t *c1, cchar_t *c2, cchar_t *hce, cchar_t *lce) {
+    static cchar_t space = {
+        .attr = A_REVERSE,
+        .chars = {' ', '\0'}
+    };
+    cchar_t c1r = *c1, c2r = *c2;
+    c1r.attr |= A_REVERSE;
+    c2r.attr |= A_REVERSE;
     if(l1 > l2) {
-        mvvline(ph+1-l1, x, c1, l1-l2 );
-        mvvline(ph+1-l2, x, c2|A_REVERSE, l2 );
+        mvvline_set(ph+1-l1, x, c1, l1-l2 );
+        mvvline_set(ph+1-l2, x, &c2r, l2 );
     } else if(l1 < l2) {
-        mvvline(ph+1-l2, x, (c2==hce || c2==lce) ? c2|A_REVERSE : ' '|A_REVERSE,  l2-l1 );
-        mvvline(ph+1-l1, x, c1|A_REVERSE, l1 );
+        mvvline_set(ph+1-l2, x, (c2==hce || c2==lce) ? &c2r : &space,  l2-l1 );
+        mvvline_set(ph+1-l1, x, &c2r, l1 );
     } else {
-        mvvline(ph+1-l2, x, c2|A_REVERSE, l2 );
+        mvvline_set(ph+1-l2, x, &c2r, l2 );
     }
 }
 
-void plot_values(int ph, int pw, double *v1, double *v2, double max, double min, int n, chtype pc, chtype hce, chtype lce, double hm) {
+void plot_values(int ph, int pw, double *v1, double *v2, double max, double min, int n, cchar_t *pc, cchar_t *hce, cchar_t *lce, double hm) {
     const int first_col=3;
     int i=(n+1)%pw;
     int x;
@@ -199,7 +208,7 @@ void paint_plot(void) {
     asctime_r(lt, ls);
     mvaddstr(height-2, width-strlen(ls), ls);
 
-    mvvline(height-2, 5, plotchar|A_NORMAL, 1);
+    mvvline_set(height-2, 5, &plotchar, 1);
     mvprintw(height-2, 7, "last=%.1f min=%.1f max=%.1f avg=%.1f %s ",  values1[n], min1, max1, avg1, unit);
     if(rate)
         printw(" interval=%llds", (long long int)td);
@@ -209,7 +218,7 @@ void paint_plot(void) {
         mvprintw(height-1, 7, "last=%.1f min=%.1f max=%.1f avg=%.1f %s   ",  values2[n], min2, max2, avg2, unit);
     }
 
-    plot_values(plotheight, plotwidth, values1, values2, max, hardmin, n, plotchar, max_errchar, min_errchar, hardmax);
+    plot_values(plotheight, plotwidth, values1, values2, max, hardmin, n, &plotchar, &max_errchar, &min_errchar, hardmax);
 
     draw_axes(height, plotheight, plotwidth, max, hardmin, unit);
 
@@ -251,9 +260,13 @@ int main(int argc, char *argv[]) {
     int show_ver;
     int show_usage;
 
-    plotchar=T_VLINE;
-    max_errchar='e';
-    min_errchar='v';
+    setlocale(LC_ALL, "");
+    if (MB_CUR_MAX > 1)            // if non-ASCII characters are supprted:
+        plotchar.chars[0]=0x2502;  // U+2502 box drawings light vertical
+    else
+        plotchar.chars[0]='|';     // U+007C vertical line
+    max_errchar.chars[0]='e';
+    min_errchar.chars[0]='v';
 
     cached_opterr = opterr;
     opterr=0;
@@ -303,16 +316,16 @@ int main(int argc, char *argv[]) {
                 break;
             case '2':
                 two=1;
-                plotchar='|';
+                plotchar.chars[0]='|';
                 break;
             case 'c':
-                plotchar=optarg[0];
+                mbtowc(&plotchar.chars[0], optarg, MB_CUR_MAX);
                 break;
             case 'e':
-                max_errchar=optarg[0];
+                mbtowc(&max_errchar.chars[0], optarg, MB_CUR_MAX);
                 break;
             case 'E':
-                min_errchar=optarg[0];
+                mbtowc(&min_errchar.chars[0], optarg, MB_CUR_MAX);
                 break;
             case 's':
                 softmax=atof(optarg);
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -31,17 +31,15 @@
     #define VERSION_STR STR(VERSION_MAJOR) "." STR(VERSION_MINOR) "." STR(VERSION_PATCH)
 #endif
 
+#define T_RARR '>'
+#define T_UARR '^'
 #ifdef NOACS
 #define T_HLINE '-'
 #define T_VLINE '|'
-#define T_RARR '>'
-#define T_UARR '^'
 #define T_LLCR 'L'
 #else
 #define T_HLINE ACS_HLINE
 #define T_VLINE ACS_VLINE
-#define T_RARR ACS_RARROW
-#define T_UARR ACS_UARROW
 #define T_LLCR ACS_LLCORNER
 #endif
 
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -5,7 +5,6 @@
 // Apache License 2.0
 //
 
-#define _XOPEN_SOURCE 500  // Get ncurses wchar_t support from SUSv2 (UNIX 98)
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>


hartwork:
Hi @edgar-bonet, very nice! Compiled and ran on Gentoo. Two small things:

I get this warning when compiling:

```console
# make
cc -Wall -Wextra    ttyplot.c  `pkg-config --libs ncursesw 2>/dev/null || echo '-lcursesw -ltinfo'` -o ttyplot
ttyplot.c: In function ‘draw_line’:
ttyplot.c:135:9: warning: implicit declaration of function ‘mvvline_set’; did you mean ‘mvvline’? [-Wimplicit-function-declaration]
  135 |         mvvline_set(ph+1-l1, x, c1, l1-l2 );
      |         ^~~~~~~~~~~
      |         mvvline
```
This is with GCC 12.

The other thing is that the `^` and `>` on the axes' end look "less ASCII" now. Maybe it's just my font but I liked before better personally:

##### Before
before_Screenshot_20231113_150724



##### After
after_Screenshot_20231113_150755

edgar-bonet:
@hartwork: Thanks for testing!

> `implicit declaration of function ‘mvvline_set’`

You may be using the wrong include file, or the wrong feature-test macro. On my Ubuntu, /usr/include/curses.h has this:

```c
#ifndef NCURSES_WIDECHAR
#if defined(_XOPEN_SOURCE_EXTENDED) || (defined(_XOPEN_SOURCE) && (_XOPEN_SOURCE - 0 >= 500))
#define NCURSES_WIDECHAR 1
#else
#define NCURSES_WIDECHAR 0
#endif
#endif /* NCURSES_WIDECHAR */
```

It then declares all wide character functions (including `mvvline_set()`) if `NCURSES_WIDECHAR` is non-zero.

Could you check how it works on your Gentoo? Is `mvvline_set()` defined in a different file? Is its definition conditioned on something else? Would it be appropriate to directly define `NCURSES_WIDECHAR`? Does `pkg-config --cflags ncursesw` give some good clue?

> The other thing is that the ^ and > on the axes' end look "less ASCII" now. Maybe it's just my font but I liked before better personally

I also like the “before” better. I'll have to check why this happened...

hartwork:
> ```c
> #ifndef NCURSES_WIDECHAR
> #if defined(_XOPEN_SOURCE_EXTENDED) || (defined(_XOPEN_SOURCE) && (_XOPEN_SOURCE - 0 >= 500))
> #define NCURSES_WIDECHAR 1
> #else
> #define NCURSES_WIDECHAR 0
> #endif
> #endif /* NCURSES_WIDECHAR */
> ```

@edgar-bonet same here.

> It then declares all wide character functions (including `mvvline_set()`) if `NCURSES_WIDECHAR` is non-zero.
> 
> Could you check how it works on your Gentoo? Is `mvvline_set()` defined in a different file?

A different file, yes:
```console
# fgrep -R mvvline_set /usr/include/ncurses*
/usr/include/ncursestw/ncurses.h:extern NCURSES_EXPORT(int) mvvline_set (int, int, const cchar_t *, int);       /* generated:WIDEC */
/usr/include/ncursestw/ncurses.h:#define mvvline_set(y,x,c,n)           mvwvline_set(stdscr,(y),(x),(c),(n))
/usr/include/ncursestw/curses.h:extern NCURSES_EXPORT(int) mvvline_set (int, int, const cchar_t *, int);        /* generated:WIDEC */
/usr/include/ncursestw/curses.h:#define mvvline_set(y,x,c,n)            mvwvline_set(stdscr,(y),(x),(c),(n))
/usr/include/ncursesw/ncurses.h:extern NCURSES_EXPORT(int) mvvline_set (int, int, const cchar_t *, int);        /* generated:WIDEC */
/usr/include/ncursesw/ncurses.h:#define mvvline_set(y,x,c,n)            mvwvline_set(stdscr,(y),(x),(c),(n))
/usr/include/ncursesw/curses.h:extern NCURSES_EXPORT(int) mvvline_set (int, int, const cchar_t *, int); /* generated:WIDEC */
/usr/include/ncursesw/curses.h:#define mvvline_set(y,x,c,n)             mvwvline_set(stdscr,(y),(x),(c),(n))
```

None of these seem to be pulled in from `curses.h` automatically…

```console
$ grep '#.*include' /usr/include/curses.h | sort
#include <libutf8.h>
#include <ncurses_dll.h>
#include <stdarg.h>     /* we need va_list */
#include <stdbool.h>
#include <stddef.h>     /* we want wchar_t */
#include <stdint.h>
#include <stdio.h>
#include <stdnoreturn.h>
#include <unctrl.h>
#include <wchar.h>              /* ...to get mbstate_t, etc. */
```
…so it may need an additional include.

> Does `pkg-config --cflags ncursesw` give some good clue?

It does, good idea!
```console
# pkg-config --cflags ncursesw
-D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 -I/usr/include/ncursesw
```

So the `Makefile` would need respect these flags also and then it may work out of the box.

> > The other thing is that the ^ and > on the axes' end look "less ASCII" now. Maybe it's just my font but I liked before better personally
> 
> I also like the “before” better. I'll have to check why this happened...

My guess is that that is what `ncursesw` makes of the `ACS_*` values used in the code.

hartwork (changes requested):
@edgar-bonet so it would be great to have the build system add `pkg-config --cflags ncursesw` to CFLAGS. PS: The pull request title has a small typo "mu[l]tibyte".

edgar-bonet:
@hartwork wrote:
> so it would be great to have the build system add `pkg-config --cflags ncursesw` to CFLAGS

I just pushed a commit with this change. Does it fix the warning you see?

hartwork (commented):
> I just pushed a commit with this change. Does it fix the warning you see?

@edgar-bonet yes it does, thank you!

edgar-bonet:
About the axes ends changing from `^`/`>` to `↑`/`→`, @hartwork wrote:
> My guess is that that is what `ncursesw` makes of the `ACS_*` values used in the code.

Indeed. It appears that two conditions are needed for this to happen:

* link with ncursesw
* call `setlocale()` with a locale supporting `↑`/`→`.

One way to avoid this change would be to unconditionally define `T_*ARR` to `'^'`/`'>'`. Or maybe make them wide characters too and use either `L'▲'`/`L'▶'` or `L'△'`/`L'▷'`.

@tenox7: Do your vintage Unix systems support wide characters? Specifically, is `mbtowc()` provided? Do they have a version of the curses library that provides `mvvline_set()`? If not, either this will have to be a build option, or we may try @MIvanchev's idea of outputting multibyte strings directly.

hartwork:
> One way to avoid this change would be to unconditionally define `T_*ARR` to `'^'`/`'>'`. Or maybe make them wide characters too and use either `L'▲'`/`L'▶'` or `L'△'`/`L'▷'`.

@edgar-bonet :+1:

edgar-bonet:
As this pull request's branch conflicted with master, I rebased and force-pushed.

@hartwork, could you please test this on Gentoo? I am worried about the include file: the previous version of this pull request added /usr/include/ncursesw to the include path, whereas commit 45e0db87bbc6caaf1d33f805247f70e16af542aa (which landed in master today) changed the header name to ncurses.h. Rebasing the PR on master changes the include file again:

```
this PR (old):  /usr/include/curses.h  → /usr/include/ncursesw/curses.h
commit 45e0db8: /usr/include/curses.h  → /usr/include/ncurses.h
this PR (new):  /usr/include/ncurses.h → /usr/include/ncursesw/ncurses.h
```

Also, is the last commit's message correct? It says

> On Gentoo, however, /usr/include/ncurses.h and /usr/include/ncursesw/ncurses.h are different files, and only the latter defines the wide character functions

tenox7:
feel free to change it to ncursesw or whatever is needed for wide chars, I just wanted to break away from old curses; or revert my change if you think `curses.h` is better

edgar-bonet:
@tenox7: For me, it makes no difference at all: on Ubuntu, the files
* /usr/include/ncurses.h
* /usr/include/ncursesw/curses.h
* /usr/include/ncursesw/ncurses.h

are all symlinks to /usr/include/curses.h. That file declares everything we may need: the classic curses API, the ncurses extensions and – given the proper `#define` – the wide character extensions.

The relevant question is about portability: What would be the most portable way to get the declarations we want? Are there systems where `ncurses.h` is needed to get the ncurses or wchar extensions? Are there systems that lack `ncurses.h` but have everything in `curses.h`? Where should we search for those files? `pkg-config` seems like a good way to get the include paths but, if we want to support systems without `pkg-config` (do we want that?), what fallback path should we use? I do not have the answers to those questions. :frowning:

FWIW, the Debian package ncurses-doc contains man pages that recommend including `curses.h`, and a HOWTO that recommends `ncurses.h` instead. :confused: The ncurses package also came with two `pkg-config` package definitions: one with and one without the wchar extensions. They differ only in the name and the `-l` compiler option:

```diff
--- a/usr/lib/x86_64-linux-gnu/pkgconfig/ncurses.pc
+++ b/usr/lib/x86_64-linux-gnu/pkgconfig/ncursesw.pc
@@ -1,19 +1,19 @@
 # pkg-config file generated by gen-pkgconfig
 # vile:makemode
 
 prefix=/usr
 exec_prefix=${prefix}
 libdir=/usr/lib/x86_64-linux-gnu
 includedir=${prefix}/include
 abi_version=6
 major_version=6
 version=6.3.20211021
 
-Name: ncurses
+Name: ncursesw
 Description: ncurses 6.3 library
 Version: ${version}
 URL: https://invisible-island.net/ncurses
 Requires.private: 
-Libs:  -L/usr/lib/x86_64-linux-gnu -lncurses -ltinfo
+Libs:  -L/usr/lib/x86_64-linux-gnu -lncursesw -ltinfo
 Libs.private:  -ldl 
 Cflags:  -D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600
```

Makefile
```diff
@@ -2,7 +2,8 @@ DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
 CFLAGS += -Wall -Wextra
-LDLIBS += `pkg-config --libs ncurses 2>/dev/null || echo '-lcurses -ltinfo'`
+CFLAGS += `pkg-config --cflags ncursesw 2>/dev/null || echo '-D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 -I/usr/include/ncursesw'`
+LDLIBS += `pkg-config --libs ncursesw 2>/dev/null || echo '-lcursesw -ltinfo'`
```
hartwork, line 6: @edgar-bonet I believe the `-ltinfo` may need to be `-ltinfow` with `w` also bease on:

```
# pkgconf --libs ncursesw
-lncursesw -ltinfow
```

Commit 5bdee13933ba65cef945f19ab1a3c998140b9feb compiles well on Gentoo :+1: 

Yes, it's two different files with regard to content:

```
# ls -l /usr/include/{ncursesw/,}{,n}curses.h
-rw-r--r-- 1 root root  82280 Apr 13  2023 /usr/include/curses.h
lrwxrwxrwx 1 root root      8 Apr 13  2023 /usr/include/ncurses.h -> curses.h
-rw-r--r-- 1 root root 100387 Apr 13  2023 /usr/include/ncursesw/curses.h
lrwxrwxrwx 1 root root      8 Apr 13  2023 /usr/include/ncursesw/ncurses.h -> curses.h
```

ttyplot.c
```diff
@@ -258,9 +266,13 @@ int main(int argc, char *argv[]) {
...
-    plotchar=T_VLINE;
-    max_errchar='e';
-    min_errchar='v';
+    setlocale(LC_ALL, "");
+    if (MB_CUR_MAX > 1)            // if non-ASCII characters are supprted:
+        plotchar.chars[0]=0x2502;  // U+2502 box drawings light vertical
+    else
+        plotchar.chars[0]='|';     // U+007C vertical line
+    max_errchar.chars[0]='e';
+    min_errchar.chars[0]='v';
```
hartwork, line 275: @edgar-bonet I think we may need `memset(x, sizeof(x), 0)` for all three of these prior to setting `.chars[0]` to ensure that the other bytes are also initialized.

hartwork (changes requested):

hartwork:
> The relevant question is about portability: What would be the most portable way to get the declarations we want? Are there systems where `ncurses.h` is needed to get the ncurses or wchar extensions? Are there systems that lack `ncurses.h` but have everything in `curses.h`? Where should we search for those files? `pkg-config` seems like a good way to get the include paths but, if we want to support systems without `pkg-config` (do we want that?), what fallback path should we use? I do not have the answers to those questions. 😦

@edgar-bonet my vote would be for requiring widely available `ncurses` (https://repology.org/project/ncurses/versions) and command `pkg-config` (sometimes provided by pkgconf). That would allow dropping the `|| echo ...` (that has a hard time working for e.g. Gentoo and Debian at the same, as we learned today, further up). @tenox7 what do you think?

Makefile
edgar-bonet: I see. It's `-ltinfo` on Ubuntu and `-ltinfow` on Gentoo. Should I change the fallback to `-ltinfow`? If, as you suggest in a further comment, we drop the `pkg-config` fallbacks (presumably in another PR), then I can leave this as-is for now.

PS: thanks for testing and for the info on Gentoo's headers.

ttyplot.c
edgar-bonet: `plotchar`, `max_errchar` and `min_errchar` are statically-allocated globals. They are zeroed-out by the run-time at program startup.

edgar-bonet:
@hartwork: This PR would not work with traditional curses: it relies on the wchar extensions, which come with ncurses. The thing is, we can usually import the ncurses API by including either `curses.h` or `ncurses.h`, and I don't know whether one option is more portable than the other.

Regarding `pkg-config`, I wonder how many systems would be left out by such requirement. According to Wikipedia, it is available for Linux, BSD, Windows, macOS, and Solaris. Does not look like an unreasonable requirement to me.

ttyplot.c
hartwork: @edgar-bonet are you sure that global variables *without* keyword `static` — our case here — have "static storage duration"? I found these two related pieces in https://www.open-std.org/jtc1/sc22/wg14/www/docs/n1256.pdf , draft version of C99:

This:
static_automatic_Screenshot_20231117_151545

And then (in **6.7.8**(.10)):

c99_again_Screenshot_20231117_151722

Isn't 6.2.4.**4** our case? it would be indeterminate according to 6.7.8.10 above then, or at least the compiler would have a right to, correct?

hartwork:
> @hartwork: This PR would not work with traditional curses: it relies on the wchar extensions, which come with ncurses. The thing is, we can usually import the ncurses API by including either `curses.h` or `ncurses.h`, and I don't know whether one option is more portable than the other.

@edgar-bonet I'd be happy either way, maybe with a slight preference for `ncurses.h` because it says what we developed against and doesn't pretend to support something else as of today. We could look at more distros if we find any with ncurses that does not install both files but I'd doubt we'd find any. One thing we could do is see how much of a chance the new code has to work with [pdcurses](https://pdcurses.org/) either way but only supporting *n*curses would probably be acceptable. My personal interest in pdcurses is limited by the fact that neither Debian nor Gentoo have a package for it, which makes it feel a bit too niche to me right now, but [some distros have pdcurses packaged](https://repology.org/project/pdcurses/versions). How do you feel about the whole topic (not just pdcurses)?

> Regarding `pkg-config`, I wonder how many systems would be left out by such requirement. According to Wikipedia, it is available for Linux, BSD, Windows, macOS, and Solaris. Does not look like an unreasonable requirement to me.

:+1:

ttyplot.c
MIvanchev: 6.2.4.3 is the case because the linkage is per default external for file scope objects.

hartwork:
> @hartwork: I'm happy with including `ncurses.h`.
>
> I hadn't heard about PDCurses before. Seems niche indeed: it's described as “_A curses library for environments that don't fit the termcap/terminfo model_” (not what we target), it doesn't provide a Makefile for Unix other than X11 and SDL, and most distro packages describe it as “Curses for Windows/MinGW/X11”. I would say don't bother unless someone volunteers to maintain the port.

@edgar-bonet :+1: 

> Long term, it would be nice to have an easy way to run tests on multiple Linux and non-Linux OSes. A Vagrantfile could be an option if/when a list of officially supported systems is established.

https://github.com/mpartel/bindfs/commits/vagrant-ci has a work-in-progress CI using Vagrant based on GitHub Actions, but it's a bit slow whenever a CI runner does not come with KVM acceleration, there seems to be some roulette about it. Free alternatives for running Vagrant inside some CI may include AppVeyor and Cirrus CI, but it would need to be tried/verified. Also Vagrant images for all target OSes and the right driver would be needed, e.g. the VirtualBox Vagrant images only work if the driver is VirtualBox, same for KVM/Qemu.

tenox7:
yeah lets make ncurses a requirement, I'm happy with that; ncurses are available even for very old unixes; I haven't seen or used pdcurses in a very very long time

edgar-bonet:
@tenox7: Did you have a chance to take look at this PR? As far an I am concerned, it is ready, but I would happily amend it if requested.

Should I change the arrows at the end of the axes (`↑`/`→`) within this PR? If so, should I use `^`/`>`, `▲`/`▶` or `△`/`▷`?

tenox7:
sorry, for slow responses, let me check

tenox7:
I want to test it a little bit in different OSes and terminals.

hartwork:
> Should I change the arrows at the end of the axes (`↑`/`→`) within this PR? If so, should I use `^`/`>`, `▲`/`▶` or `△`/`▷`?

@edgar-bonet my vote for ASCII `^` and `>` or those, yes please, the current arrows are not too sexy. @tenox7 how do you feel about the topic?

edgar-bonet:
I pushed a commit that unconditionally sets the arrow tips to the ASCII characters `^`/`>`.

hartwork:
@edgar-bonet good call :+1:

tenox7:
It doesn't build on macOS 

```
cc -Wall -Wextra -lcursesw ttyplot.c -o ttyplot
ttyplot.c:47:1: error: unknown type name 'cchar_t'; did you mean 'wchar_t'?
cchar_t plotchar, max_errchar, min_errchar;
^~~~~~~
wchar_t
```
(repeated many times)

tenox7:
yes, just make on macos

```
$ git clone https://github.com/edgar-bonet/ttyplot.git && cd ttyplot && git checkout multibyte
$ make
cc -Wall -Wextra `pkg-config --cflags ncursesw 2>/dev/null || echo '-D_DEFAULT_SOURCE -D_XOPEN_SOURCE=600 -I/usr/include/ncursesw'`    ttyplot.c  `pkg-config --libs ncursesw 2>/dev/null || echo '-lcursesw -ltinfo'` -o ttyplot
ttyplot.c:47:1: error: unknown type name 'cchar_t'; did you mean 'wchar_t'?
cchar_t plotchar, max_errchar, min_errchar;
^~~~~~~
wchar_t
```

```
$ pkg-config --cflags ncursesw
-D_DARWIN_C_SOURCE
```

edgar-bonet:
@tenox7 wrote:
> yes, just make on macos

There is something fishy going on: `pkg-config` does find the package `ncursesw` (where the `w` stands for “widechar”), yet the CFLAGS it gives cannot get us the widechar API. :-(

Could you try to see how many files matching the glob `*curses*.h` you have on your system? Do any of these define the type `cchar_t`? If so, is the definition protected by some `#if` or `#ifdef`? If not, does any of your `*.h` files define `cchar_t`?

PS: As there was a conflict, I rebased on master and force-pushed.

tenox7:
I think there may be a conflict between curses from SDK and Homebrew. Checking.

tenox7:
I did this and it worked:

```
cc -Wall -Wextra -D_DARWIN_C_SOURCE -D_XOPEN_SOURCE_EXTENDED    -o ttyplot ttyplot.c -lncurses
```

edgar-bonet:
@tenox7: Thanks for debugging the issue!

I checked `man ncurses`, and it says `_XOPEN_SOURCE_EXTENDED` is deprecated in favor of `_XOPEN_SOURCE`. I tried the later, for consistency with what `pkg-config` sets on Linux, but the CI build failed on Mac. So I just followed your suggestion, and CI is happy. :smile:

hartwork (approved):
I could not be happier that the new macOS CI from #108 could be of help here. Let's get this merged :+1:

PR #101, opened 2023-11-15
Remove code repetition from `plot_values()`

edgar-bonet:
This is just for making the code DRYer, and thus easier to refactor. No changes in the program behavior.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -134,20 +134,13 @@ void draw_line(int x, int ph, int l1, int l2, chtype c1, chtype c2, chtype hce, 
 }
 
 void plot_values(int ph, int pw, double *v1, double *v2, double max, double min, int n, chtype pc, chtype hce, chtype lce, double hm) {
-    int i;
-    int x=3;
+    const int first_col=3;
+    int i=(n+1)%pw;
+    int x;
     max-=min;
 
-    for(i=n+1; i<pw; i++)
-        draw_line(x++, ph,
-                  (v1[i]>hm) ? ph  : (v1[i]<min) ?  1  : (int)(((v1[i]-min)/max)*(double)ph),
-                  (v2[i]>hm) ? ph  : (v2[i]<min) ?  1  : (int)(((v2[i]-min)/max)*(double)ph),
-                  (v1[i]>hm) ? hce : (v1[i]<min) ? lce : pc,
-                  (v2[i]>hm) ? hce : (v2[i]<min) ? lce : pc,
-                  hce, lce);
-
-    for(i=0; i<=n; i++)
-        draw_line(x++, ph,
+    for(x=first_col; x<first_col+pw; x++, i=(i+1)%pw)
+        draw_line(x, ph,
                   (v1[i]>hm) ? ph  : (v1[i]<min) ?  1  : (int)(((v1[i]-min)/max)*(double)ph),
                   (v2[i]>hm) ? ph  : (v2[i]<min) ?  1  : (int)(((v2[i]-min)/max)*(double)ph),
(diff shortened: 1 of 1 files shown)


ttyplot.c
```diff
@@ -134,20 +134,12 @@ void draw_line(int x, int ph, int l1, int l2, chtype c1, chtype c2, chtype hce,
...
-        draw_line(x++, ph,
-                  (v1[i]>hm) ? ph  : (v1[i]<min) ?  1  : (int)(((v1[i]-min)/max)*(double)ph),
-                  (v2[i]>hm) ? ph  : (v2[i]<min) ?  1  : (int)(((v2[i]-min)/max)*(double)ph),
-                  (v1[i]>hm) ? hce : (v1[i]<min) ? lce : pc,
-                  (v2[i]>hm) ? hce : (v2[i]<min) ? lce : pc,
-                  hce, lce);
-
-    for(i=0; i<=n; i++)
-        draw_line(x++, ph,
+    for(x=3; x<3+pw; x++, i=(i+1)%pw)
```
hartwork, line 141: Can we we move `x=3` back to the declaration to give an empty initializer? Just an idea.

The `x<3+pw` is not yet clear to me. If it's correct it could use a comment, if not let's fix it :smiley: What's the idea?

edgar-bonet: We could move `x=3` to the initializer, but I am afraid that would make the `for` loop less clear. The whole point of the `for` construct is to put group all the loop-control statements together, at the beginning of the loop. Actually, I would prefer to put the declaration within the `for` statement, in C99-style: `for(int x=3; ...`. I only refrained to do so because it would clash with the mostly-C90 style of the current code base.

`x` is the x coordinate in the window, i.e. the column number on the terminal. The plot starts at column 3, and `pw` is its width, thus the condition `x<3+pw`. Would it be clearer if written like below?

```c
int n;
for(n=0; n<pw; n++, i=(i+1)%pw)
        draw_line(n+3, ph, ...);
```

hartwork: > We could move `x=3` to the initializer, but I am afraid that would make the `for` loop less clear. The whole point of the `for` construct is to put group all the loop-control statements together, at the beginning of the loop. Actually, I would prefer to put the declaration within the `for` statement, in C99-style: `for(int x=3; ...`. I only refrained to do so because it would clash with the mostly-C90 style of the current code base.

@edgar-bonet I see. Personally I believe status quo is the worst of the three options. I'd always put the assignment into the declaration even if that means it's outside the loop for non-C99. Nevermind, I'm not a gate here, I'm glad to now understand your reasoning, thanks!

> `x` is the x coordinate in the window, i.e. the column number on the terminal. The plot starts at column 3, and `pw` is its width, thus the condition `x<3+pw`.

An offset, of course! Sorry :smile: 

> Would it be clearer if written like below?
> 
> ```c
> int n;
> for(n=0; n<pw; n++, i=(i+1)%pw)
>         draw_line(n+3, ph, ...);
> ```

It may be, but maybe it's just me and I'm biased now that you explained it to me. The increase in number of additions is probably not measurable, maybe it is easier to understand, maybe `const int x_offset = 3` and using that constant would even add more ease. But I'm happy now, please go forward as you consider best.

edgar-bonet: I went for the constant holding the column offset. Squashed and force-pushed.

hartwork (approved):
@edgar-bonet I like it, let's see what  @tenox7 thinks.

ttyplot.c
hartwork: @edgar-bonet thanks! :+1:

tenox7 (approved):
LGTM

tenox7:
thank you very much!

PR #103, opened 2023-11-16
Potentially incorrect usage of `optind`.

### LINKED ISSUE #102: arguments broken
paging @MIvanchev https://github.com/tenox7/ttyplot/pull/90 broke aruments/flags (update: for bsd/macos only), can you please fix?

MIvanchev:
This should fix #102.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -23,7 +23,7 @@
 #define STR(x) STR_(x)
 #define VERSION_MAJOR 1
 #define VERSION_MINOR 5
-#define VERSION_PATCH 1
+#define VERSION_PATCH 2
 #if VERSION_PATCH == 0
     #define VERSION_STR STR(VERSION_MAJOR) "." STR(VERSION_MINOR)
 #else
@@ -301,8 +301,14 @@ int main(int argc, char *argv[]) {
         exit(0);
     }
 
-    // Reset the getopt index as per spec.
-    optind = 0;
+    // Run a 2nd iteration over the arguments to actually process the options.
+    // According to getopt's documentation this is done by setting optind to 1
+    // (or 0 in some special cases). On BSDs and Macs optreset must be set to 1
+    // in addition.
+    optind = 1;
+#if defined(__OpenBSD__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__DragonFly__) || defined(__APPLE__)
+    optreset = 1;
+#endif
 
     while((c=getopt(argc, argv, optstring)) != -1) {
         switch(c) {
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,5 +1,5 @@
 name: ttyplot
-version: 1.5.1
+version: 1.5.2
 summary: realtime plotting utility for terminal/console with data input from stdin
 description: |
   takes data from standard input / unix pipeline,


hartwork (changes requested):
@MIvanchev can we replace "Reset the getopt index as per spec." with a sentence on *why* `optind` needs a reset at all? My understanding is that it needs a reset because we want to do a recsan or we would ignore some options by mistake?

MIvanchev:
@hartwork good idea!

hartwork:
@MIvanchev thanks for the adjustment!

I found this sentence in `man optind`:

> A program that scans multiple argument vectors, or rescans the same vector more than once, and wants to make use of GNU extensions such as '+' and '-' at the start of optstring, or  changes  the value of POSIXLY_CORRECT between scans, must reinitialize getopt() by resetting optind to 0, rather than the traditional value of 1.  (Resetting to 0 forces the invocation of an internal initialization routine that rechecks POSIXLY_CORRECT and checks for GNU extensions in optstring.)

Can you confirm that we do _not_ need/have any of that?

hartwork:
> I don't think we need, but it seems like MacOS' getopt has different semantics regarding `optind`: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getopt.3.html

What difference did you observe in detail?

I notice now that "rescans the same vector more than once" is exactly our case. 

Do we need one version of the code for GNU getopt and one version for non-GNU maybe? Just thinking aloud.

hartwork:
@MIvanchev PS: Also thinking aloud: maybe the most portable fix would be to merge the two scans, to avoid rescanning. What was the key idea behind separating the two?

MIvanchev:
>  What was the key idea behind separating the two scans?

To avoid doing anything, including `scanf` and `printf`, if the user has requested `-v` or `-h` and to have a well-defined order for `-v` and `-h`.

hartwork:
> > I don't think we need, but it seems like MacOS' getopt has different semantics regarding `optind`: https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/getopt.3.html
> 
> What difference did you observe in detail?

I found this now in the macOS man page that you linked above:

> In order to use `getopt()` to evaluate multiple sets of arguments, or to evaluate a single set of arguments multiple times, the variable `optreset` must be set to 1 before the second and each additional set of calls to `getopt()`, and the variable `optind` must be reinitialized.

I see.

hartwork:
> > What was the key idea behind separating the two scans?
> 
> To avoid doing anything, including `scanf` and `printf`, if the user has requested `-v` or `-h`

@MIvanchev so an optimization rather than a strong need, okay.

> and to have a well-defined order for `-v` and `-h`.

How would the order not be well-defined without? What am I missing?

ttyplot.c
```diff
@@ -301,8 +301,14 @@ int main(int argc, char *argv[]) {
...
-    // Reset the getopt index as per spec.
-    optind = 0;
+    // Run a 2nd iteration over the arguments to actually process the options.
+    // According to getopt's documentation this is done by setting optind to 1
+    // (or 0 in some special cases). On BSDs and Macs optreset must be set to 1
+    // in addition.
+    optind = 1;
+#if defined(__OpenBSD__) || (defined(__APPLE__) && defined(__MACH__))
+    optreset = 1;
+#endif
```
hartwork, line 311: - What's the idea about `__MACH__` here?
- Are we sure yet that (or when) `optind` 1 is fine for glibc?
- Here's how dnsdb solves the same problem differently: https://github.com/dnsdb/dnsdbq/blob/a3b77edda4b4f66d27e8b12c2c551c8dd1c2dcd4/dnsdbq.c#L1181-L1196

tenox7: @MIvanchev yeah could you add ifs for glibc/nonglibc, I like how dnsdbq does it

we should test on some non-glibc linux as well

hartwork: PS: This is where glibc's getopt is doing additonal work for `optind` 0: https://github.com/bminor/glibc/blob/b7f8b6b64b135e6dba5083d688675874d6809c91/posix/getopt.c#L482

hartwork: > we should test on some non-glibc linux as well

Alpine has musl, seems like a good candidate

MIvanchev: I do not find **any** indication that glibc mandates 0 vs 1 so I believe dnsdbq is cargo culting in this case. However it might make sense to check for `__FreeBSD__` / `__NetBSD__`.

MIvanchev: @hartwork 
> What's the idea about __MACH__ here?
IMHO __APPLE__ and __MACH__ together indicate a MacOS installation. I don't think we need to check further than that but I'm not a Mac user so open for suggestions.

> Are we sure yet that (or when) optind 1 is fine for glibc?
Yes.

hartwork: > I do not find **any** indication that glibc mandates 0 vs 1 so I believe dnsdbq is cargo culting in this case.

@MIvanchev did you see my link to glibc code above?:

> PS: This is where glibc's getopt is doing additonal work for `optind` 0: https://github.com/bminor/glibc/blob/b7f8b6b64b135e6dba5083d688675874d6809c91/posix/getopt.c#L482

MIvanchev: Yes I did, it does additional work for the 0 case but we do not require the additional work as the documentation states. No extensions are used and `POSIXLY_CORRECT` is not changed in-between.

hartwork: > IMHO **APPLE** and **MACH** together indicate a MacOS installation.

@MIvanchev if https://sourceforge.net/p/predef/wiki/OperatingSystems/#macos is correct — and they usually are — then:
- It's macOS X specifically, with emphasis on X.
- Only set by GNU C and the Intel compiler.

Unless we have strong need for `__MACH__` that I don#t yet see, I'd vote for dropping `__MACH__`.

Actually I think I'd vote to exact-copy dnsdb's solution. PS: I read that MirBSD seems to also define the one for OpenBSD so it should be covered for free.

hartwork: > Yes I did, it does additional work for the 0 case but we do not require the additional work as the documentation states. No extensions are used and `POSIXLY_CORRECT` is not changed in-between.

@MIvanchev cool, thanks! :+1:

MIvanchev: I dropped `__MACH__` but I'm leaving `optind = 1`. dnsdb is really just paranoid here for no reason at all.

hartwork: @MIvanchev their approach is arguably more robust and can be copy pasted for re-use elsewhere even to places where extensions are used, and they code will survive if someone adds an extension without noticing. Maybe they don't need it today, but paranoid being negative in connotation don't see them doing anything wrong, rather the contrary. Just my 2 cents.

MIvanchev: You're really overthinking this IMO, just because someone did something somewhere doesn't mean it's a good idea. Cargo cult or false interpretation of the documentation is not robust programming. This `__MACH__` stuff was a different beast and really increases robustness.

Regarding your other question about the well-defined order for `-v` and `-h`. If you have:

```
ttyplot -v -h
ttyplot -h -v
ttyplot -v -v -v -h
```

One should always have a fixed precedence over the other.

edgar-bonet:
The second scan of the options does nothing other than set some variables, call `atof()` and `snprintf()`. There is no harm in doing this if the user asks for `-v` or `-h`. Merging the two scans would not only completely remove this issue, it would also make the code simpler.

ttyplot.c
hartwork: > You're really overthinking this IMO, just because someone did something somewhere doesn't mean it's a good idea.

@MIvanchev I don't need to be right about anything here, I just like to understand things better and be sure we fix a good way. I'm not sure where you read me say anything like that, it doesn't sound like what I intended to say or said literally.

> Cargo cult or false interpretation of the documentation is not robust programming.

Not sure how that relates.

> The code will survive anyhow because the second you add extensions it won't work without `= 0` and you'll just change it. 

Not work somewhere != not work everywhere consistently. It may go unnoticed in a project with Linux-only CI for instance, e.g. ttyplot.

> This `__MACH__` stuff was a different beast and really increases robustness.

I don't see how. But before either of us has access to actual macOS 9 we're probably both guessing.

> Regarding your other question about the well-defined order for `-v` and `-h`. If you have:
> 
> ```
> ttyplot -v -h
> ttyplot -h -v
> ttyplot -v -v -v -h
> ```
> 
> One should always have a fixed precedence over the other.

How/why is a fixed preference more important than order with help and version?

ttyplot.c
```diff
@@ -301,8 +301,14 @@ int main(int argc, char *argv[]) {
...
     }
 
-    // Reset the getopt index as per spec.
-    optind = 0;
+    // Run a 2nd iteration over the arguments to actually process the options.
+    // According to getopt's documentation this is done by setting optind to 1
+    // (or 0 in some special cases). On BSDs and Macs optreset must be set to 1
+    // in addition.
+    optind = 1;
+#if defined(__OpenBSD__) || defined(__FreeBSD__) || defined(__NetBSD__) || defined(__APPLE__)
```
hartwork, line 309: I think we're about to discriminate DragonFly BSD here *unless*…
- DragenFly BSD also defines one of the other macros? (which I haven't seen indication for so far)
- We add [`__DragonFly__`](https://sourceforge.net/p/predef/wiki/OperatingSystems/#dragonfly) to the mix

What do you think?

hartwork (changes requested):

edgar-bonet:
@MIvanchev: I bet merging the two scans would solve the problem with _far less_ comments than the current approach.

ttyplot.c
MIvanchev: It was about the dnsdbq crew, not you, I understand and value your motivation :)

>How/why is a fixed preference more important than order with help and version?

I mean that the program does the same with `-v -h` as with `-h -v` and the order doesn't matter.

ttyplot.c
MIvanchev: `__DragonFly__` check added.

hartwork (changes requested):
@MIvanchev thanks for the adjustments! Could you squash the 7 commits together and do a force push so that we keep clean and to-the-point history on `master` even if merged with a merge commit? I'd hit "approve" here then, thanks! :+1:

hartwork (approved):
@MIvanchev thanks! :+1:

tenox7:
Tested on FreeBSD and MacOS. LGTM. Thank you!!!!!!

PR #104, opened 2023-11-16
torture.c: Cover compilation in CI + fix signed integer overflow undefined behavior

hartwork:
CC @edgar-bonet


edgar-bonet (approved):
LGTM.

PR #106, opened 2023-11-18
Increase the time resolution in rate mode

edgar-bonet:
The program torture.c outputs a waveform of amplitude&nbsp;5 and period 0.72&nbsp;s. Its peak derivative is therefore expected to be

    5 × (2π ÷ 0.72) ≈ 43.6

However,

```shell
./torture | ./ttyplot -r
```

shows a waveform that peaks at about&nbsp;0.5, which is off by two orders of magnitude. The reason is that data points arrive every 10&nbsp;ms, and ttyplot “rounds” this value to a full second, effectively underestimating the rate by a factor&nbsp;100.

This pull request fixes this behavior by handling time with microsecond resolution. It also moves the computation of rates to its own function for better code clarity.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -11,6 +11,7 @@
 #include <unistd.h>
 #include <float.h>
 #include <time.h>
+#include <sys/time.h>
 #include <ncurses.h>
 #include <signal.h>
 #include <errno.h>
@@ -46,14 +47,13 @@
 
 sigset_t sigmsk;
 chtype plotchar, max_errchar, min_errchar;
-time_t t1,t2,td;
+struct timeval now;
+double td;
 struct tm *lt;
 double max=FLT_MIN;
 double softmax=FLT_MIN, hardmax=FLT_MAX, hardmin=0.0;
 char title[256]=".: ttyplot :.", unit[64]={0}, ls[256]={0};
 double values1[1024]={0}, values2[1024]={0};
-double cval1=FLT_MAX, pval1=FLT_MAX;
-double cval2=FLT_MAX, pval2=FLT_MAX;
 double min1=FLT_MAX, max1=FLT_MIN, avg1=0;
 double min2=FLT_MAX, max2=FLT_MIN, avg2=0;
 int width=0, height=0, n=0, r=0, v=0, c=0, rate=0, two=0, plotwidth=0, plotheight=0;
@@ -77,6 +77,35 @@ void usage(void) {
 
 void version(void) {
     printf("ttyplot %s\n", VERSION_STR);
+}
+
+// Replace *v1 and *v2 (if non-NULL) by their time derivatives.
+//  - v1, v2: addresses of input data and storage for results
+//  - now: current time
+// Return time since previous call.
+double derivative(double *v1, double *v2, const struct timeval *now)
+{
+    static double previous_v1, previous_v2, previous_t = DBL_MAX;
+    const double t = now->tv_sec + 1e-6 * now->tv_usec;
+    const double dt = t - previous_t;
+    previous_t = t;
+    if (v1) {
+        const double dv1 = *v1 - previous_v1;
+        previous_v1 = *v1;
+        if (dt <= 0)
+            *v1 = 0;
+        else
+            *v1 = dv1 / dt;
+    }
+    if (v2) {
+        const double dv2 = *v2 - previous_v2;
+        previous_v2 = *v2;
+        if (dt <= 0)
+            *v2 = 0;
+        else
+            *v2 = dv2 / dt;
+    }
+    return dt;
 }
 
 void getminmax(int pw, double *values, double *min, double *max, double *avg, int v) {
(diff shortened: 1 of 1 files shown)


hartwork (commented):
Hi @edgar-bonet I think there is a chance that there is a misunderstanding of what `-r` was meant to do. Maybe I'm wrong, that's okay, let's find out!

The way I understand the current `-r` is that it is meant to be used with accumulators, values that only ever go up. For example, let's take a data source like this:

```console
# while grep numa_hit /proc/vmstat | awk '{print $2}' ; do sleep 0.25 ; done
113308336
113308915
113309576
113310145
113310700
113311263
113311825
113312415
113312971
113313544
^C
```

Now if we diff these values, we can see they are roughly 700 apart:
```console
# a=0; while sleep 0.25; do b="$(grep numa_hit /proc/vmstat | awk '{print $2}')"; echo "$((b - a))"; a="${b}"; done
113417603
678
685
688
686
692
782
706
693
695
698
^C
```

When I pipe that as input to `./ttyplot -r` now as…

```console
# while grep numa_hit /proc/vmstat | awk '{print $2}' ; do sleep 0.25 ; done | ./ttyplot -r
```

…on `master`, ttyplot display values and average match that "roughly 700 apart" reality, seems like a match.

When I do the same now on branch `rate`(commit 0863f67adb2c33440e7018af6d0fca213bdf2c10) the values and average are way higher, more like 2400.

Therefore my impression is that we have *two different scenarios here* and that maybe we need an additional flag to support your scenario *in addition* under a new ciommand line argument rather than replacing the current `-r`.

What do you think?

edgar-bonet:
@hartwork: Replace `sleep 0.25` with `sleep 2`. You will see that both master an this PR's branch compute the same thing: a rate of roughly 2400&nbsp;hits/s.

hartwork:
> @hartwork: Replace `sleep 0.25` with `sleep 2`.

@edgar-bonet … because the old code uses a timer with seconds resolution and therefore 0.25 is too fast for it then I suppose, interesting!

I guess then maybe this pull request is just doing the right thing? :smiley: 

The variable naming `x` and `y` seem a bit unfortunate given that (1) both seem to be y coordinates and (2) the application uses x and y already for 2-dimensional coordinates in ncurses space. If you could pick a different pair of names, I'd likely be all for the new ones. What do you think?

ttyplot.c
```diff
@@ -79,6 +79,35 @@ void version(void) {
...
 
+// Replace *x and *y (if non-NULL) by their time derivatives.
+//  - x, y: addresses of input data and storage for results
+//  - now: current time
+// Return time since previous call.
+double derivative(double *x, double *y, const struct timeval *now)
+{
+    static double previous_x, previous_y, previous_t = DBL_MAX;
+    double t = now->tv_sec + 1e-6 * now->tv_usec;
+    double dt = t - previous_t;
```
hartwork, line 90: - (I'd personally add two `const` here. Same for `dx` and `dy` below. Just an idea.)

edgar-bonet:
@hartwork wrote:
> The variable naming `x` and `y` seem a bit unfortunate

Indeed it is. I replaced them with `v1` and `v2`, which are more consistent with the current naming. I also added `const` qualifiers as you suggested. Squashed and force-pushed.

hartwork (approved):
@edgar-bonet thanks for the adjustments :+1: Let's see how @tenox7 likes it.

tenox7:
interesting, so basically if samples for rate calculator are sub second interval the data will be lost?

tenox7:
also note that torture.c was never designed for rate testing, it would be cool if you could add it as well, maybe make the sleep/interval an arg/flag and option for rate (-r) ?

edgar-bonet:
@tenox7 wrote:
> if samples for rate calculator are sub second interval the data will be lost?

Not exactly. Each number received on the input adds one bar to the plot, so no data is lost. The issue is that, in master, the values are incorrect. If the interval is large, the rates displayed are real rates (i.e. time derivatives), in counts per second. If the interval becomes smaller than one second, then `ttyplot` “rounds” it up to one full second, which results in overestimated rates.

The issue also impacts intervals that are longer (but not much longer) than one second. For example, the command

```shell
for n in $(seq 1 10 1000); do echo $n; sleep 1.5; done
```

creates a counter that increments by ten units every 1.5&nbsp;seconds. This is a rate of 6.7&nbsp;counts per second. Piping this to `ttyplot` in master shows a rate that alternates between 5 and 10. With this pull request, `ttyplot` shows a very steady rate of 6.7.

> torture.c was never designed for rate testing, it would be cool if you could add it as well [...]

I would gladly do so. Within this PR or in a follow-up one?

hartwork:
> I would gladly do so. Within this PR or in a follow-up one?

My vote for merging to `master` what we have here and _then_ doing a follow-up. Fewer bits in flight, baby steps, shorter-lived branches. Clear vote for a merge from my side.

PR #109, opened 2023-11-20
Makefile: Make command pkg-config a hard build dependency

hartwork:
Note that the idea behind `which pkg-config` is to get a *crystal clear single error message* rather than a long list of link errors.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
 CFLAGS += -Wall -Wextra
-LDLIBS += `pkg-config --libs ncurses 2>/dev/null || echo '-lcurses -ltinfo'`
+LDLIBS += `pkg-config --libs ncurses`
 torture: LDLIBS = -lm
 
 all: ttyplot
@@ -18,6 +18,13 @@ uninstall:
 	rm -f $(DESTDIR)$(MANPREFIX)/man1/ttyplot.1
 
 clean:
-	rm -f ttyplot torture
+	rm -f ttyplot torture *.o
 
-.PHONY: all clean install uninstall
+require_pkgconfig:
+	which pkg-config
+
+ttyplot.o: require_pkgconfig
+
+torture.o: require_pkgconfig
+
+.PHONY: all clean install uninstall require_pkgconfig


tenox7 (approved):
LGTM

tenox7:
thank you!

Makefile
```diff
@@ -18,6 +18,13 @@ uninstall:
 	rm -f $(DESTDIR)$(MANPREFIX)/man1/ttyplot.1
 
 clean:
-	rm -f ttyplot torture
+	rm -f ttyplot torture *.o
 
-.PHONY: all clean install uninstall
+require_pkgconfig:
+	which pkg-config
```
eli-schwartz, line 24: > Note that the idea behind `which pkg-config` is to get a _crystal clear single error message_ rather than a long list of link errors.

Sorry, this is broken and doesn't work on my machine. `which` is a somewhat uncommon command to have installed, I don't have it installed, and there are a couple different implementations of it floating about in the wild -- not all of them even reliably return an error when a command doesn't exist. Some of them print "no such command" and return success!

For users that don't have `which` installed, this just causes the build to always, always fail -- even when pkg-config exists, and even when ncurses exists.

If it was just a matter of avoiding the erroneous dependency on `which`, I would suggest:

```suggestion
	command -v pkg-config
```

...

But the significantly worse *conceptual* problem that leads to the fundamental intention of this PR to be broken and incomplete and just all-around fail to work... is that the foundational premise is that ncurses is a required dependency, but you added a hard dependency on pkg-config and which, instead of pkg-config and ncurses.

So as it is, it seems like what you really want is:
```suggestion
	pkg-config --exists ncurses
```

eli-schwartz: For some in-depth discussions on the topic, see:

- https://mywiki.wooledge.org/BashFAQ/081
- https://unix.stackexchange.com/questions/85249/why-not-use-which-what-to-use-then/85250#85250

I'm afraid to say that `which` is just really bad to use all around. The main reason why it is often used is just because Debian's policy manual mandates that it be installed as part of the base system, so that debian maintainers can use it in maintainer scripts -- which means that it is available by default on the single most common distro. But it is a portability problem everywhere else. :(

hartwork: > Sorry, this is broken and doesn't work on my machine. `which` is a somewhat uncommon command to have installed, I don't have it installed[.]

@eli-schwartz could you add which operating system you are running that does not have `which` installed?

edgar-bonet: @hartwork: Yes, go ahead with `command -v`. In the next couple of days I may be less responsive than I would like to.

eli-schwartz: > > Sorry, this is broken and doesn't work on my machine. `which` is a somewhat uncommon command to have installed, I don't have it installed[.]
> 
> @eli-schwartz could you add which operating system you are running that does not have `which` installed?

Arch Linux and Gentoo, from personal experience -- but there are various others. e.g. Void Linux doesn't have it installed by default, and OpenSUSE has a `which` command implemented as a shell function in /usr/etc/profile.d/alias.bash that is available on the interactive console but NOT in a Makefile or shell script.

hartwork: @eli-schwartz thanks for elaborating :+1: 

PS: I'm on Gentoo myself and a Gentoo box without proper `which` is unlikely because it's part of `@system` (the core set of packages for other readers):

```console
# eix --system --only-names | grep which
sys-apps/which
```

I'll get this fixed, thanks for the report :+1:

hartwork: > @hartwork: Yes, go ahead with `command -v`. In the next couple of days I may be less responsive than I would like to.

@edgar-bonet good, thanks for the heads up :+1:

eli-schwartz: > PS: I'm on Gentoo myself and a Gentoo box without proper `which` is unlikely because it's part of `@system` (the core set of packages for other readers):

https://bugs.gentoo.org/646588

```
$ cat /etc/portage/profile/packages
-*sys-apps/which
```

As I said, I don't have it installed.

That doesn't mean all Gentoo boxes don't have it installed, which is why I listed 3 other distros where the general populace is not guaranteed to have it installed.

In the future, Gentoo will also probably not have it installed ;) and packages that insist on it anyway will end up having to add it to BDEPEND / RDEPEND as needed.

hartwork: > > PS: I'm on Gentoo myself and a Gentoo box without proper `which` is unlikely because it's part of `@system` (the core set of packages for other readers):
> 
> https://bugs.gentoo.org/646588

@eli-schwartz thanks for sharing, that's interesting!

> ```
> $ cat /etc/portage/profile/packages
> -*sys-apps/which
> ```
> 
> As I said, I don't have it installed.

Before https://bugs.gentoo.org/646588 is fixed, Gentoo without `which` is not Gentoo. Let's fix it for the other distros and future Gentoo though.

PR #110, opened 2023-11-21
Handle keyboard input

edgar-bonet:
The normal way to stop ttyplot is by typing <kbd>Ctrl</kbd>-<kbd>C</kbd>. This is unfortunate, as this key combination is normally only ever used for interrupting an unresponsive program: well behaved programs can be terminated by some regular (not signal) keyboard command.

This pull request makes ttyplot quit when the user hits <kbd>q</kbd> on the keyboard. This is the same key used to terminate more, less, htop, ncdu, and other full-screen terminal programs. Once we monitor keyboard input, adding extra keybindings is cheap. Then, as a bonus, the <kbd>r</kbd> key toggles rate mode on and off.

Keyboard commands only work when stdin is redirected (our most common use case), as otherwise keyboard input is interpreted as data. It works by reading the keystrokes from the controlling terminal (`/dev/tty`).

This pull request depends on PR #98, which creates the required framework for concurrently watching multiple event sources.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -19,6 +19,9 @@
 #include <ncurses.h>
 #include <signal.h>
 #include <errno.h>
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
 
 #ifdef __OpenBSD__
 #include <err.h>
@@ -363,6 +366,16 @@ int main(int argc, char *argv[]) {
 
     redraw_screen(errstr);
 
+    // If stdin is redirected, open the terminal for reading user's keystrokes.
+    int tty = -1;
+    if (!isatty(STDIN_FILENO))
+        tty = open("/dev/tty", O_RDONLY);
+    if (tty != -1) {
+        // Disable input line buffering. The function below works even when stdin
+        // is redirected: it searches for a terminal in stdout and stderr.
+        cbreak();
+    }
+
     sigemptyset(&empty_sigset);
 
     sigemptyset(&block_sigset);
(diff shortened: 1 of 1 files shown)


ttyplot.c
```diff
@@ -370,6 +373,17 @@ int main(int argc, char *argv[]) {
...
 
+    // If stdin is redirected, open the terminal for reading user's keystrokes.
+    int tty = -1;
+    if (!isatty(STDIN_FILENO))
+        tty = open("/dev/tty", O_RDONLY);
+    if (tty != -1) {
+        // Disable input line buffering and echo. The functions below work even when
+        // stdin is redirected: they search for a terminal in stdout and stderr.
+        cbreak();
+        noecho();
```
hartwork, line 384: There is a call to `noecho();` already a few lines up in line 368. Does the first call render the second unecessary or is there a reason to have a second call that I am missing?

hartwork (approved):
@edgar-bonet neat! :+1: 

I was thinking yesterday or so that that hitting a key — say <kbd>p</kbd> — to **p**ause the animation when you saw something and thought "wait, pause, what was that?". Would be nice as a follow-up pull request if @tenox7 liked the general direction.

ttyplot.c
edgar-bonet: It's me: I did not notice the previous call, despite it being pretty close... Fixed, force-pushed.

hartwork: @edgar-bonet thanks!

MIvanchev:
I mean this MR augments half of the program. Is it really necessary to have bells and whistles? Personally I prefer ^-C for termination and it's handled quite well by ttyplot.

edgar-bonet:
@MIvanchev wrote:
> I mean this MR augments half of the program.

What do you mean by “augments half of the program”? This pull request adds only 28 lines of code (+6 comment-only and blank lines).

Please, beware that this PR depends on PR #98. If you click on the “Files changed” tab on GitHub, you will see the total diff from the merge base. You have to manually select (with Shift-click) the three last commits in order to see this PR alone.

edgar-bonet:
Rebased on top of the updated PR #98.

hartwork:
> You have to manually select (with Shift-click) the three last commits in order to see this PR alone.

This compare view URL would also work (as longs as it remains to be exactly three commits and the branch exists): https://github.com/edgar-bonet/ttyplot/compare/keys~3...keys

PR #114, opened 2023-11-24
Respect plot char for case `ttyplot -c X -2` (fixes #113)

### LINKED ISSUE #113: [1.5.2] `ttyplot -c X -2` fails to use ASCII character "X" for drawing?
Hi!

I ran into something by accident that to me looks like there could be more issues with the command line parsing than just #102 that was fixed by #103 recently. This is what I found:

```console
# git rev-parse HEAD
07ce2f904f0a24ab46c0af170fa65cbb3bfe62bb

# uname -s
Linux

# make

# ./ttyplot -2 -c X <<<'1 2'  # good, draws using the "X" character as expected

# ./ttyplot -c X -2 <<<'1 2'  # bad, draws using default pipe ("|") character, no error reported
```

Do you get the same? Am I using it wrong?

Thanks, Sebastian

hartwork:
Fixes #113

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -354,7 +354,6 @@ int main(int argc, char *argv[]) {
                 break;
             case '2':
                 two=1;
-                plotchar.chars[0]='|';
                 break;
             case 'c':
                 mbtowc(&plotchar.chars[0], optarg, MB_CUR_MAX);


edgar-bonet (approved):
LGTM

PR #115, opened 2023-11-24
Refactor the handling of input data

edgar-bonet:
**Draft pull request**

The purpose of this pull request is to simplify the non-blocking handling of input data. The relevant code is taken out of the main loop and split into three small functions. The main loop now strictly follows this simple pattern:

```c
while (1) {
    wait_for_next_event();
    handle_the_events_we_got();
}
```

The data events are fully handled within the loop iteration that caught them, even if they provide more than one data record. The data is now tokenized and parsed using `strtok()`/`strtod()` instead of `sscanf()`. Tokens that do not parse to numbers are interpreted as zeroes: no NANs are generated.

This PR is marked as “draft” because it is intended to be merged in a future “development” branch, that integrates previous pull requests. A such, it sits on top of my branch “next”, which merges PRs #98, #99, #106, #110 and #114. Note that, since NANs are not generated, PR #112 becomes irrelevant to this branch. This PR proper is a single commit that shrinks ttyplot.c by 40&nbsp;lines. See [the diff][].

[the diff]: /edgar-bonet/ttyplot/compare/next..refactor-input

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -69,6 +69,7 @@ double min1=FLT_MAX, max1=FLT_MIN, avg1=0;
 double min2=FLT_MAX, max2=FLT_MIN, avg2=0;
 int width=0, height=0, n=-1, r=0, v=0, c=0, rate=0, two=0, plotwidth=0, plotheight=0;
 bool fake_clock = false;
+char *errstr = NULL;
 const char *verstring = "https://github.com/tenox7/ttyplot " VERSION_STR;
 
 void usage(void) {
@@ -322,9 +323,115 @@ void redraw_screen(const char * errstr) {
     refresh();
 }
 
+// Return a pointer to the last occurrence within [s, s+n) of one of the bytes in the string accept.
+char *find_last(char *s, size_t n, const char *accept)
+{
+    for (int pos = n - 1; pos >= 0; pos--) {
+        if (strchr(accept, s[pos]))
+            return s + pos;
+    }
+    return NULL;  // not found
+}
+
+// Handle a single value from the input stream.
+// Return whether we got a full data record.
+bool handle_value(double value)
+{
+    static double saved_value;
+    static int saved_value_valid = 0;
+
+    // First value of a 2-value record: save it for later.
+    if (two && !saved_value_valid) {
+        saved_value = value;
+        saved_value_valid = 1;
+        return false;
+    }
+
+    // Otherwise we have a full record.
+    n = (n+1) % plotwidth;
+    if (two) {
+        values1[n] = saved_value;
+        values2[n] = value;
+        saved_value_valid = 0;
+    } else {
+        values1[n] = value;
+    }
+    if (rate)
+        td = derivative(&values1[n], two ? &values2[n] : NULL, &now);
+    return true;
+}
+
+// Handle a chunk of input data: extract the numbers, store them, redraw if needed.
+// Return the number of bytes consumed.
+size_t handle_input_data(char *buffer, size_t length)
+{
+    static const char delimiters[] = " \t\r\n";  // white space
+
+    // Find the last delimiter.
+    char *end = find_last(buffer, length, delimiters);
+    if (!end)
+        return 0;
+    *end = '\0';
+
+    // Tokenize and parse.
+    int records = 0;  // number or records found
+    char *str = buffer;
+    char *token;
+    while ((token = strtok(str, delimiters)) != NULL) {
+        char *number_end;
+        double value = strtod(token, &number_end);
+        if (*number_end != '\0')  // garbage found
+            value = 0;
+        if (handle_value(value))
+            records++;
+        str = NULL;  // tell strtok() to stay on the same string
+    }
+    v += records;
+    if (records > 0)
+        redraw_screen(errstr);
+    return end - buffer + 1;
+}
+
+// Handle an "input ready" event, where only a single read() is guaranteed to not block.
+// Return whether the input stream got closed.
+bool handle_input_event(void)
+{
+    static char buffer[4096];
+    static size_t buffer_pos = 0;
+
+    // Buffer incoming data.
+    ssize_t bytes_read = read(STDIN_FILENO, buffer + buffer_pos, sizeof(buffer) - 1 - buffer_pos);
+    if (bytes_read < 0) {  // read error
+        if (errno == EINTR || errno == EAGAIN)  // we should try again later
+            return false;
+        errstr = strerror(errno);  // other errors are considered fatal
+        return true;
+    }
+    if (bytes_read == 0) {
+        errstr = "input stream closed";
+        buffer[buffer_pos++] = '\n';  // attempt to extract one last value
+        handle_input_data(buffer, buffer_pos);
+        redraw_screen(errstr);  // redraw to display the error message
+        return true;
+    }
+    buffer_pos += bytes_read;
+
+    // Handle this new data.
+    size_t bytes_consumed = handle_input_data(buffer, buffer_pos);
+
+    // If we have excessive garbage, discard a bunch. This is to ensure that we can always ask read for >= 1K bytes,
+    // and keep good performance, especially with high input pressure.
+    if (buffer_pos - bytes_consumed > sizeof(buffer) / 2)
+        bytes_consumed += sizeof(buffer) / 4;
+
+    if (bytes_consumed > 0 && bytes_consumed < buffer_pos)
+        memmove(buffer, buffer + bytes_consumed, buffer_pos - bytes_consumed);
+    buffer_pos -= bytes_consumed;
+    return false;
+}
+
 int main(int argc, char *argv[]) {
     int i;
-    char *errstr = NULL;
     bool stdin_is_open = true;
     int cached_opterr;
     const char *optstring = "2rc:e:E:s:m:M:t:u:vh";
@@ -466,27 +573,7 @@ int main(int argc, char *argv[]) {
     signal(SIGWINCH, resize);
     signal(SIGINT, finish);
 
-    char input_buf[4096] = "";
-    size_t input_len = 0;
-
     while(1) {
-        if (sigint_pending) {
-            break;
-        }
-
-        if (sigwinch_pending) {
-            sigwinch_pending = 0;
-
-            endwin();
-            initscr();
-            clear();
-            refresh();
-
-            gethw();
-
-            goto redraw_and_continue;
-        }
-
         // Block until (a) we receive a signal or (b) stdin can be read without blocking
         // or (c) timeout expires, in order to reduce use of CPU and power while idle
         fd_set read_fds;
@@ -501,20 +588,27 @@ int main(int argc, char *argv[]) {
             if (tty >= select_nfds)
                 select_nfds = tty + 1;
         }
-        const bool previous_parse_succeeded = (r == (two ? 2 : 1));
-        struct timespec timeout;
-        timeout.tv_sec = 0;
-        if (previous_parse_succeeded) {
-            timeout.tv_nsec = 0;  // we may have more input pressing, let's not throttle it down
-        } else {
-            timeout.tv_nsec = 500 * 1000 * 1000;  // <=500 milliseconds for a healthy clock display
-        }
+        struct timespec timeout = { .tv_sec = 0, .tv_nsec = 500e6 };  // 500 milliseconds for refreshing the clock
         const int select_ret = pselect_without_signal_starvation(select_nfds, &read_fds, NULL, NULL, &timeout, &empty_sigset);
 
-        const bool signal_received = ((select_ret == -1) && (errno == EINTR));
+        gettimeofday(&now, NULL);
 
-        if (signal_received) {
-            continue;  // i.e. skip right to signal handling
+        // Refresh the clock on timeouts.
+        if (select_ret == 0)
+            redraw_screen(errstr);
+
+        // Handle signals.
+        if (sigint_pending) {
+            break;
+        }
+        if (sigwinch_pending) {
+            sigwinch_pending = 0;
+            endwin();
+            initscr();
+            erase();
+            refresh();
+            gethw();
+            redraw_screen(errstr);
         }
 
         // Handle user's keystrokes.
@@ -532,131 +626,14 @@ int main(int argc, char *argv[]) {
             }
         }
 
-        const bool stdin_can_be_read_without_blocking = ((select_ret > 0) && FD_ISSET(STDIN_FILENO, &read_fds));
-
-        // Read as much from stdin as we can (first read after select is non-blocking)
-        if (stdin_can_be_read_without_blocking) {
-            // Last resort: truncate existing input if input line ever is
-            //              too long
-            if (input_len >= sizeof(input_buf) - 1 /* terminator */) {
-                input_len = 0;
-            }
-
-            char * const read_target = input_buf + input_len;
-            const size_t max_bytes_to_read = sizeof(input_buf) - (input_len + 1 /* terminator */);
-            const ssize_t bytes_read_or_error = read(STDIN_FILENO, read_target, max_bytes_to_read);
-
-            if (bytes_read_or_error > 0) {
-                input_len += bytes_read_or_error;
-                assert(input_len < sizeof(input_buf));
-                input_buf[input_len] = '\0';
-            } else {
-                assert(bytes_read_or_error <= 0);
-                if (bytes_read_or_error == 0) {
-                    close(STDIN_FILENO);
-                    errstr = "input stream closed";
-                    stdin_is_open = false;
-                } else {
-                    assert(bytes_read_or_error == -1);
-                    if ((errno != EINTR) && (errno != EAGAIN)) {
-                        errstr = strerror(errno);
-                        stdin_is_open = false;
-                    }
-                }
+        // Handle input data.
+        if (select_ret > 0 && FD_ISSET(STDIN_FILENO, &read_fds)) {
+            bool input_closed = handle_input_event();
+            if (input_closed) {
+                close(STDIN_FILENO);
+                stdin_is_open = false;
             }
         }
-
-        // Extract values from record; note that the record may turn out incomplete
-        double d1 = 0.0;
-        double d2 = 0.0;
-        int record_len = -1;
-
-        if(two)
-            r = sscanf(input_buf, "%lf %lf%*[ \t\r\n]%n", &d1, &d2, &record_len);
-        else
-            r = sscanf(input_buf, "%lf%*[ \t\r\n]%n", &d1, &record_len);
-
-        // We need to detect and avoid mis-parsing "1.23" as two records "1.2" and "3"
-        const bool supposedly_complete_record = (r == (two ? 2 : 1));
-        const bool trailing_whitespace_present = (record_len != -1);
-
-        if (supposedly_complete_record && ! trailing_whitespace_present) {
-            const bool need_more_input = stdin_is_open;
-            if (need_more_input) {
-                r = 0;  // so that the parse is not mis-classified as a success further up
-                goto redraw_and_continue;
-            }
-
-            record_len = input_len;  // i.e. the whole thing
-        }
-
-        // In order to not get stuck with non-doubles garbage input forever,
-        // we need to drop input that we know(!) will never parse as doubles later.
-        if (! supposedly_complete_record && (input_len > 0)) {
-            char * walker = input_buf;
-
-            while (isspace(*walker)) walker++;  // skip leading whitespace (if any)
-
-            while ((*walker != '\0') && ! isspace(*walker)) walker++;  // skip non-double
-
-            if (two) {
-                if (*walker == '\0') {
-                    goto redraw_and_continue;
-                }
-
-                while (isspace(*walker)) walker++;  // skip gap whitespace
-
-                if (*walker == '\0') {
-                    goto redraw_and_continue;
-                }
-
-                while ((*walker != '\0') && ! isspace(*walker)) walker++;  // skip non-double
-            }
-
-            if (*walker == '\0') {
-                goto redraw_and_continue;
-            }
-
-            while (isspace(*walker)) walker++;  // skip trailing whitespace (if any)
-
-            record_len = walker - input_buf;  // i.e. how much to drop
-        }
-
-        // Drop the record that we just processed (well-formed or not) from the input buffer
-        if ((input_len > 0) && (record_len > 0)) {
-            char * move_source = input_buf + record_len;
-            const size_t bytes_to_move = input_len - record_len;
-            memmove(input_buf, move_source, bytes_to_move);
-            input_len = bytes_to_move;
-            input_buf[input_len] = '\0';
-        }
-
-        if (! supposedly_complete_record) {
-            goto redraw_and_continue;
-        }
-
-        v++;
-
-        if (n < plotwidth - 1)
-            n++;
-        else
-            n=0;
-
-        values1[n] = d1;
-        values2[n] = d2;
-
-        if(values1[n] < 0)
-            values1[n] = 0;
-        if(values2[n] < 0)
-            values2[n] = 0;
-
-        gettimeofday(&now, NULL);
-        if (rate)
-            td=derivative(&values1[n], two ? &values2[n] : NULL, &now);
-
-redraw_and_continue:
-        gettimeofday(&now, NULL);
-        redraw_screen(errstr);
     }
 
     endwin();
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -70,6 +70,7 @@ double min2=FLT_MAX, max2=FLT_MIN, avg2=0;
 int width=0, height=0, n=-1, r=0, v=0, c=0, rate=0, two=0, plotwidth=0, plotheight=0;
 bool fake_clock = false;
 char *errstr = NULL;
+bool redraw_needed = false;
 const char *verstring = "https://github.com/tenox7/ttyplot " VERSION_STR;
 
 void usage(void) {
@@ -388,7 +389,7 @@ size_t handle_input_data(char *buffer, size_t length)
     }
     v += records;
     if (records > 0)
-        redraw_screen(errstr);
+        redraw_needed = true;
     return end - buffer + 1;
 }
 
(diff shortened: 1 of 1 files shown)


ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+size_t parse_input(char *buffer, size_t length)
+{
+    static const char delimiters[] = " \t\r\n";  // white space
+
+    // Find the last delimiter.
+    int end;
+    for (end = length - 1; end >= 0; end--) {
+        if (strchr(delimiters, buffer[end]))  // found
+            break;
+    }
```
hartwork, line 331: This seems to find the rightmost delimiter. Can we leverage `strcspn` here to simplify? if not, this could be extracted as a function with a good name.

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+    if (end < 0)  // no delimiter
+        return 0;
+    buffer[end] = '\0';
+
+    // Tokenize and parse.
+    int records = 0;  // number or records found
+    char *str = buffer;
+    char *token;
+    while ((token = strtok(str, delimiters)) != NULL) {
+        records += handle_value(strtod(token, NULL));
```
hartwork, line 341: I have yet to verify that but `strtod(token, NULL)` (with empaphis on `NULL`) should tolerate some garbage around the double value and hence render half-broken input as healthy whereas `sscanf` would have discarded the token altogether. Is that change in behavior intended?

It maybe worth noting here that replacing all-garbage tokens by 0.0 is not a side effect but our explicitly wanted behavior. To a reader not knowing, the code as is looks like we're just forgetting to check for errors. Maybe that's worth a small comment in code.

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+        str = NULL;  // tell strtok() to stay on the same string
+    }
+    v += records;
+    if (records > 0)
+        redraw_screen(errstr);
```
hartwork, line 346: If `parse_input` is about parsing input, I would like to argue that it calling `redraw_screen` is out of place here because it violates the contract of what is the job of function `parse_input`. Can we pull that call out some way? I believe all calls to `redraw_screen` would ideally be from the `main` function.

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
     refresh();
 }
 
+// Handle a single value from the input stream.
+// Return whether we got a full data record.
+int handle_value(double value)
```
hartwork, line 295: We have stdbool already include, let's return `bool` for this please.

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+    }
+    v += records;
+    if (records > 0)
+        redraw_screen(errstr);
+    return end + 1;
+}
+
+// Handle an "input ready" event, where only a single read() is guaranteed to not block.
+// Return whether the input stream got closed.
+int handle_input(void)
```
hartwork, line 352: - This `int` is another `bool` in disguise. Let's use `bool` please.
- (Function `handle_input` has two jobs rather than one: (1) push stdin into the buffer, and (2) process the buffer content wrapping the call to `parse_input`. Maybe it cleaner splitting it up in two.)

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+        if (errno == EINTR || errno == EAGAIN)  // we should try again later
+            return 0;
+        errstr = strerror(errno);  // other errors are considered fatal
+        return 1;
+    }
+    if (bytes_read == 0) {
+        errstr = "input stream closed";
+        buffer[buffer_pos++] = '\n';  // attempt to extract one last value
+        parse_input(buffer, buffer_pos);
+        redraw_screen(errstr);  // redraw to display the error message
```
hartwork, line 369: (same as above)
If `handle_input` is about handling input, I would like to argue that it calling `redraw_screen` is out of place here because it violates the contract of what is the job of function `handle_input`. Can we pull that call out some way? I believe all calls to `redraw_screen` would ideally be from the `main` function.

ttyplot.c
```diff
@@ -429,29 +520,9 @@ int main(int argc, char *argv[]) {
...
-            refresh();
-
-            gethw();
-
-            goto redraw_and_continue;
-        }
-
         // Block until (a) we receive a signal or (b) stdin can be read without blocking
-        // or (c) timeout expires, in oder to reduce use of CPU and power while idle
+        // or (c) timeout expires, in order to reduce use of CPU and power while idle
```
hartwork, line 525: (Can we extract a dedicated commit for the typo?)

hartwork (changes requested):
@edgar-bonet I needed to warm up myself about this a bit at first, but now I see the value and that's in the right direction. Here are the details I found that seemed worth bringing up:

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+
+// Handle an "input ready" event, where only a single read() is guaranteed to not block.
+// Return whether the input stream got closed.
+int handle_input(void)
+{
+    static char buffer[4096];
+    static size_t buffer_pos = 0;
+
+    // Buffer incoming data.
+    int bytes_read = read(STDIN_FILENO, buffer + buffer_pos, sizeof buffer - 1 - buffer_pos);
```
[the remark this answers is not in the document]
edgar-bonet: OK, done.

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+        parse_input(buffer, buffer_pos);
+        redraw_screen(errstr);  // redraw to display the error message
+        return 1;
+    }
+    buffer_pos += bytes_read;
+
+    // Handle this new data.
+    size_t bytes_consumed = parse_input(buffer, buffer_pos);
+    if (buffer_pos - bytes_consumed > sizeof buffer / 2)
+        bytes_consumed += sizeof buffer / 4;  // discard excessive garbage
```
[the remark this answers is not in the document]
edgar-bonet: It is indeed arbitrary, as I don't have any good rule to tell what exact amount has to be discarded. As it stands, if we have an incomplete token in the buffer (i.e. not followed by white space) larger than 2&nbsp;KiB, we discard 1&nbsp;KiB.

Do you have a suggestion?

ttyplot.c
```diff
@@ -289,9 +290,99 @@ void redraw_screen(const char * errstr) {
...
+    return 1;
+}
+
+// Parse the input data. Return the number of bytes consumed.
+size_t parse_input(char *buffer, size_t length)
+{
+    static const char delimiters[] = " \t\r\n";  // white space
+
+    // Find the last delimiter.
+    int end;
```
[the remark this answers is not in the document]
edgar-bonet: It cannot: I need the variable right after the loop.

ttyplot.c
edgar-bonet: `strcspn` wouldn't work: it scans from the beginning of the string, whereas we want to find the _last_ delimiter. What we need is a version of `strpbrk` that works in reverse, scanning from the end of the string.

I hesitated to pull this out into its own function. But it does make sense to do so: that could have been a function of `string.h`, maybe named `strnrpbrk` according to the current naming convention. I made the function, but went for a different name.

ttyplot.c
edgar-bonet: Maybe `parse_input` has just a bad name. I intended for the function to handle the chunk of input data we just grabbed. This means parsing, storing the numbers in the ring buffers and, if needed, calling `redraw_screen`.

I just renamed the function to better reflect their responsibilities:
* `handle_input` → `handle_input_event`
* `parse_input` → `handle_input_data`

> I believe all calls to `redraw_screen` would ideally be from the `main` function.

I don't think they belong there. The purpose of this refactoring is to adhere to the conventional event-driven paradigm. There, the main loop does nothing other than catching and dispatching events. The event handler for the "input ready" event is responsible for _fully_ handling the event. “Fully” means that, once it's handled, we can go to sleep and wait for the next event.

That being said, I left the handling of the input being closed in the main loop, as `stdin_is_open` is local to main. I could move `redraw_screen` there too, but then `handle_input()` would need a way to communicate that a redraw is required. Maybe a global `redraw_needed` boolean? Not sure I like that better. For me, it makes sense to say that redrawing is part of the job of handling the input... if that input implies that a redraw is needed.

ttyplot.c
edgar-bonet: Oh, I didn't notice that include, and I was still in the mood of keeping with the “traditional” C style. OK, changed.

ttyplot.c
edgar-bonet: Fixed the `int` → `bool` change.

> Function `handle_input` has two jobs rather than one:

It has to fully handle the event.

> (1) push stdin into the buffer, and (2) process the buffer content wrapping the call to `parse_input`.

(3) shift the consumed data out of the buffer.

> Maybe it cleaner splitting it up in two.

That's what I did by factoring `parse_input` out of this function. I do not see how to split it further without that feeling unnatural.

ttyplot.c
edgar-bonet: Same as above.

ttyplot.c
edgar-bonet: > Can we extract a dedicated commit for the typo fix?

OK. Done. I put the typo-fixing commit first, then the big one. May I add your Co-authored-by to this second commit?

ttyplot.c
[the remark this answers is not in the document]
hartwork: @edgar-bonet thanks!

ttyplot.c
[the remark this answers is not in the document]
hartwork: @edgar-bonet you're right, nevermind then :smile:

ttyplot.c
hartwork: @edgar-bonet cool, thank you! :+1:

ttyplot.c
```diff
@@ -289,9 +290,108 @@ void redraw_screen(const char * errstr) {
     refresh();
 }
 
+// Return a pointer to the last occurrence within [s, s+n) of one of the bytes in the string accept.
+char *find_last(const char *s, size_t n, const char *accept)
+{
+    int pos;
+    for (pos = n - 1; pos >= 0; pos--) {
```
hartwork, line 297: Could you merge declaration and definition for `pos` into one?

ttyplot.c
hartwork: @edgar-bonet extracting function `find_last` was a good move, very nice, thanks!

hartwork: PS: I was too quick with `strcspn`, should have had a closer look. Maybe it was wishful thinking.

ttyplot.c
hartwork: > `strtod()` tolerates suffix garbage, but not prefix garbage. For instance, "x=42" would be refused, whereas "318mV, 35.4°C" would be parsed as {318, 35.4}. I think this is a good thing, as it could alleviate the need for preprocessing upstream.

@edgar-bonet true, but once we start accepting half-garbage there will be no way going back without breaking existing usage. How about we remain strict for now and add a command line argument for a tolerant mode later (not in the this pull request) if we find its needed? I would vote for being strict by default, for now. What do you think?

> > It maybe worth noting here that replacing all-garbage tokens by 0.0 is not a side effect but our explicitly wanted behavior.
> 
> Indeed. But maybe it would be better to just skip garbage tokens? It could make preprocessing still easier for the user. What do you think?

There's a chance that I'm misunderstanding you here, let's see.
Turning garbage into 0.0 seems needed to not mix-up channels with `-2` mode or input "1 garbage 3 garbage 5 6" would be rendered as "<1 3> <5 6>" rather than current "<1 0> <3 0> <5 6>". So skipping seems to be trouble.
Did you have something else in mind?

hartwork (changes requested):

ttyplot.c
edgar-bonet: Oh, yes, I missed that! This is a modified version of the previous implementation, where the index was needed after the loop. I did not notice that now `pos` is only needed within the loop. Fixed.

ttyplot.c
```diff
@@ -289,9 +290,108 @@ void redraw_screen(const char * errstr) {
...
+        if (strchr(accept, s[pos]))
+            return (char *) s + pos;
```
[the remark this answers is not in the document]
edgar-bonet: If I had to choose, I would take (b), but I still prefer the way it is. My reasoning for doing it this way is that this prototype best reflects the function's contract: “I promise I will not use the pointer you give me to modify your buffer. You, of course, can do whatever you want with the pointer I give back to you: it's your buffer after all”.

Furthermore, this prototype is consistent with the functions of string.h that perform similar duties: `strchr`, `strrchr`, `strpbrk` and `strstr`. They all take a `const char *` and return a `char *`.

edgar-bonet:
> once we start accepting half-garbage there will be no way going back without breaking existing usage. How about we remain strict for now and add a command line argument for a tolerant mode later (not in the this pull request) if we find its needed? I would vote for being strict by default, for now. What do you think?

I think if we later add a command-line argument for a more tolerant mode, there would be no reason to not let that mode permanently enabled. Don't you think?

ttyplot.c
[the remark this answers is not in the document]
hartwork: > If I had to choose, I would take (b), but I still prefer the way it is. My reasoning for doing it this way is that this prototype best reflects the function's contract: “I promise I will not use the pointer you give me to modify your buffer. You, of course, can do whatever you want with the pointer I give back to you: it's your buffer after all”.

@edgar-bonet I have a hard time agreeing here: If I pass in something truly consistent like literall "hello" I now get a writable pointer into the string back, that's not sound on the type system level, let's please not do that please. I'm happy with (a) or (b) but not status quo.

> Furthermore, this prototype is consistent with the functions of string.h that perform similar duties: `strchr`, `strrchr`, `strpbrk` and `strstr`. They all take a `const char *` and return a `char *`.

Const correctness is broken all over the place in the standard library, it's not a place to look for good examples on const correctness, quite the opposite. We can look at specific examples if you like and I may miss something about one or two of them, but we will find broken signatures with regard to const correctness in the standard library.

hartwork:
> > once we start accepting half-garbage there will be no way going back without breaking existing usage. How about we remain strict for now and add a command line argument for a tolerant mode later (not in the this pull request) if we find its needed? I would vote for being strict by default, for now. What do you think?
> 
> I think if we later add a command-line argument for a more tolerant mode, there would be no reason to not let that mode permanently enabled. Don't you think?

@edgar-bonet as I said, there will be no way going back without breaking existing usage. I'd be happy with all strict or all strict plus an option for tolerance. What value do you see in being tolerant by default?

edgar-bonet:
> What value do you see in being tolerant by default?

Ease of use. Never mind, I changed it to a non-tolerant mode.

I noticed there is a CI failure on the asciinema test. I don't understand what went wrong: frame number 3 is kind of repeated, with only the clock being updated. Maybe the timings are not so reliable?

I also tried `record.sh` on my box, and it failed because asciinema does not recognize the options `--cols` and `--rows`. It turns out these are not mentioned on the asciinema documentation.

ttyplot.c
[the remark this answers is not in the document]
hartwork: @edgar-bonet thanks!

hartwork:
> > What value do you see in being tolerant by default?
> 
> Ease of use. Never mind, I changed it to a non-tolerant mode.

@edgar-bonet thank you!


> I noticed there is a CI failure on the asciinema test. I don't understand what went wrong: frame number 3 is kind of repeated, with only the clock being updated. Maybe the timings are not so reliable?

I was thinking that it made sense since your code processes multiple values per call to `pselect` while current `development` only processes one but then your code should be back at the shell *faster*. Pull request #123 also has timing issues but here it seems clear: The sanitizers make the code much slower. Let me think about options for a fix, we definitively need some and I got us into this so… :smiley: 

> I also tried `record.sh` on my box, and it failed because asciinema does not recognize the options `--cols` and `--rows`. It turns out these are not mentioned on the asciinema documentation.

Interesting! I got it from this:

```console
# asciinema --version
asciinema 2.3.0

# asciinema rec --help | grep -E 'cols|rows'
                     [--cols COLS] [--rows ROWS] [-y] [-q]
  --cols COLS           override terminal columns for recorded process
  --rows ROWS           override terminal rows for recorded process
```

We could try playing with explicit values for variables `COLUMNS` and `LINES` instead and see if we get the same effect.

hartwork:
> We could try playing with explicit values for variables `COLUMNS` and `LINES` instead and see if we get the same effect.

@edgar-bonet that only got me in trouble, one way to fix it for everyone would be pull request #127.

hartwork:
> > > I noticed there is a CI failure on the asciinema test. I don't understand what went wrong: frame number 3 is kind of repeated, with only the clock being updated. Maybe the timings are not so reliable?
> > 
> > I was thinking that it made sense since your code processes multiple values per call to `pselect` while current `development` only processes one but then your code should be back at the shell _faster_. Pull request #123 also has timing issues but here it seems clear: The sanitizers make the code much slower. Let me think about options for a fix, we definitively need some and I got us into this so… 😃

@edgar-bonet I think I know know what's going on. Two things actually:
- For the sanitizers just giving them a bit more time seems to solve the problem, I fixed #123 just now accordingly. That's good new but also the boring part. 
- For the screenshots not matching the CI turns out is doing exactly what it's supposed to do — uncovering a bug that none of us humans noticed: *On current `development` the clock is broken, it's not moving.* We're sleeping more than a second total during UI testing and yet all the screenshots show time "00:00:00" — that's not right. Your changes *here* seem to fix the clock "as a side effect" and now that the clock is moving again the changes in clock produce both more and different images. Would be interesting to see where the clock broke on the way. Damn! :smiley:

Does that make sense?

hartwork:
@edgar-bonet PS: Found the regressing commit and an easy fix, new pull request #128.

hartwork:
@edgar-bonet PPS: I have five yet-to-reply to replies from you from review further up, they are not forgotten, I have five e-mails pointing to just them, replies likely coming up Monday evening.

ttyplot.c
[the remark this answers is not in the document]
hartwork: @edgar-bonet there are at least two prior approaches for comparison to this new approach (let's call it "(c)"):
- a) `master` and release 1.5.2 would skip anything malformed until they find a line feed (`\n`)
- b) `development` since merge of pull request #98 would truncate the buffer down to all-empty once it is truly full.

Personally I like (b) best: it only acts when absolutely necessary but then gives the most relief possible, which may matter with high pressure input coming in at stdin. Both (b) and (c) have the potential to cut through the middle of a single token that would turn out a well-formed double with more bytes of input from the future later. Does (b) have some quality that you wanted to avoid when going for (c) or that motivate (c) in particular?

ttyplot.c
hartwork: @edgar-bonet I hope I can concince you still that fewer duties — a smaller "scope" — for these two functions is beneficial and that calling `redraw_screen` is just outside of their ideal scope: they already do "enough".  It should not be hard to pull the drawing call out by the way, e.g.:

- For function `handle_input_data`, we could compare the value of `v` before and after the call and re-draw whenever `v_backup` differs from `v` or something like that.  Or we add a new output parameter `bool * records_written` and inspects its value, to add an alternative.  Note how name `records_written` does not mention drawing, which would be consistent with the view that drawing is not that functions' business, even in their contract with their outside world.

- For function `handle_input_event` after the call we could check if `errstr` has now become non-`NULL` to know if a redraw is needed.

I should add that currently `handle_input_event` seems  missing one more call to `redraw_screen` right after line `errstr = strerror(errno);`.  Pulling redraw out of those two functions, we may also be able to turn three calling places into a single one.  I would be very happy then :smiley:

What do you think?

edgar-bonet:
Rebased on development, added Co-authored-by Sebastian Pipping to the commit message.

ttyplot.c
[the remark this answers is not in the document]
edgar-bonet: There is one issue with approach (b). Imagine we get a very long token like `xxx...xy42` on stdin. The buffer gets full right before reading `y`. It gets discarded. On the next read, we try to parse `y42`. It is recognized as an invalid token. This is good. Now, imagine the buffer gets full right _after_ reading `y`. On the next read, we parse `42` as a valid token. This is bad.

The rationale of approach (c) is:

* It ensures that, after processing some input, there is always 2K of free space left in the buffer. This ensures that `read` is never called with a small `count` parameter, which would be bad in case of high input pressure.

* It makes very hard for a partial invalid token to be mistakenly parsed as a valid one.

Yes, there are still edge cases that could be mishandled, like a 2K-long `12.000...000` token. I do not see any easy way of handling this king of token, and I would not get of of my way to accomodate them.

ttyplot.c
edgar-bonet: Not that I'm fan of it but... I moved the redraw to a single place of the main loop.

edgar-bonet:
The CI failure seems spurious to me: a timing issue again.

ttyplot.c
hartwork: @edgar-bonet the increase in global state is not ideal and this way drawing is still among those functions businesses… but turning more global state into passing things around is maybe better a topic for another time. Thanks for this adjustment.

hartwork:
> Rebased on development, added Co-authored-by Sebastian Pipping to the commit message.

@edgar-bonet thanks!

> The CI failure seems spurious to me: a timing issue again.

I'll see what I can do.

ttyplot.c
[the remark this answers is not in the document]
hartwork: > The rationale of approach (c) is:
> 
> * It ensures that, after processing some input, there is always 2K of free space left in the buffer. This ensures that `read` is never called with a small `count` parameter, which would be bad in case of high input pressure.

@edgar-bonet so rather than requesting [0..4k] bytes per `read` call we would request [1k..4k] bytes now, interesting. My vote for briefly documenting that motivation as a comment in code. Why only discard `sizeof(buffer) / 4` rather than full `sizeof(buffer) / 2` when air is getting thin? If there is no particular reason, `/ 2` seems more natural and would not need an additional comment. It would be [2k..4k] for all requests then, which would support the case of high pressure.

> There is one issue with approach (b). Imagine we get a very long token like `xxx...xy42` on stdin. The buffer gets full right before reading `y`. It gets discarded. On the next read, we try to parse `y42`. It is recognized as an invalid token. This is good. Now, imagine the buffer gets full right _after_ reading `y`. On the next read, we parse `42` as a valid token. This is bad.
> 
> The rationale of approach (c) is:
> 
> * [..]
> 
>  * It makes very hard for a partial invalid token to be mistakenly parsed as a valid one.

The cutting still happens by position rather than content, the chances of cutting right though `y42` should be the same — am I missing something?

> Yes, there are still edge cases that could be mishandled, like a 2K-long `12.000...000` token. I do not see any easy way of handling this king of token, and I would not get of of my way to accomodate them.

We'd need a dynamically growing buffer for that. I'm with you that we don't need that complexity. PS: If current 4096 seems too low, I think we'd be free to pick 8192 also.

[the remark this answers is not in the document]
edgar-bonet: > Why only discard `sizeof(buffer) / 4` rather than full `sizeof(buffer) / 2` when air is getting thin?

In order to avoid turning `xxx...xy42` into `42`.

> the chances of cutting right though y42 should be the same

There would be 1K right after the `42`, within the same token. The chances of this token being valid are slim.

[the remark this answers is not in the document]
edgar-bonet: > My vote for briefly documenting that motivation as a comment in code.

Too tired to come up with something. Any suggestion?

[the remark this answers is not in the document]
hartwork: > > My vote for briefly documenting that motivation as a comment in code.
> 
> Too tired to come up with something. Any suggestion?

@edgar-bonet "Ensure that we can always ask `read` for >=2k bytes and not have that number go down near zero, for better performance, especially with high pressure input?" Something like that?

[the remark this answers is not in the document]
hartwork: > > Why only discard `sizeof(buffer) / 4` rather than full `sizeof(buffer) / 2` when air is getting thin?
> 
> In order to avoid turning `xxx...xy42` into `42`.
> 
> > the chances of cutting right though y42 should be the same
> 
> There would be 1K right after the `42`, within the same token. The chances of this token being valid are slim.

@edgar-bonet I think I understand now: at the moment where…

> `buffer_pos - bytes_consumed > sizeof(buffer) / 2`

…is checked, we know that all these bytes failed to make even a single valid token — which probably means that there is no whitespace in there anywhere, correct? — and so we cut off 1k of a block of 2k that only later could have become a >=2k single token. Does that sound right? Then this approach seems to indeed offer a bonus. Definitely worth commenting in code for the next generation on this code base.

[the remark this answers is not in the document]
edgar-bonet: Thanks! Added a slight variation on that. Would you mind rebasing?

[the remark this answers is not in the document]
hartwork: @edgar-bonet I have pushed a rebased copy into branch `refactor-input-rebased` in here right now. Only the first commit had conflicts to be resolved. Please check if everything looks like expected and if so please free to replace the pull request content by `refactor-input-rebased` via reset hard and a forced pushed.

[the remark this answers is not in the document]
hartwork: @edgar-bonet the CI likes it: https://github.com/tenox7/ttyplot/actions?query=branch%3Arefactor-input-rebased

ttyplot.c
```diff
@@ -479,20 +568,25 @@ int main(int argc, char *argv[]) {
...
-        if (signal_received) {
-            continue;  // i.e. skip right to signal handling
+        // Handle signals.
+        if (sigint_pending) {
+            break;
+        }
+        if (sigwinch_pending) {
+            sigwinch_pending = 0;
+            endwin();
+            initscr();
```
hartwork, line 587: @edgar-bonet I played with resizing the window down to vertical below-minimum… — we cannot drop `erase` and `refresh` here but that's what the pull request is doing, and it would be regression. Please:

```diff
diff --git a/ttyplot.c b/ttyplot.c
index 0e824e6..cc3eba4 100644
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -607,6 +607,8 @@ int main(int argc, char *argv[]) {
             sigwinch_pending = 0;
             endwin();
             initscr();
+            erase();
+            refresh();
             gethw();
             redraw_needed = true;
         }
```

hartwork (changes requested):

ttyplot.c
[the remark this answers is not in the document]
edgar-bonet: Thank-you ! That was helpful. I took your rebased branch, restored `erase()` and `refresh()`, squashed a few commits together, and pushed again.

ttyplot.c
edgar-bonet: Good catch! Fixed, thanks.

hartwork (approved):
@edgar-bonet let's get this merged! :+1:

PR #116, opened 2023-11-24
Add options to stresstest.c

edgar-bonet:
**Draft pull request**

This pull request adds option parsing, and multiple options to torture.c, which are intended to exercise corner cases in `ttyplot`:

```shell
$ ./torture -h
Usage: ./torture [-h] [-2] [-c] [-g] [-r rate]
  -h       print this help message and exit
  -2       output two waves
  -c       randomly chunk the output
  -g       occasionally output garbage
  -r rate  sample rate in samples/s (default: 100)
```

This PR is marked as “draft” because it is intended to be merged in a future “development” branch, that integrates previous pull requests. A such, it sits on top of my branch “next”, which merges PRs #98, #99, #106, #110 and #114. This PR proper is four commits: see [the diff][].

[the diff]: /edgar-bonet/ttyplot/compare/next..torture-opts

### DIFF
--- a/stresstest.c
+++ b/stresstest.c
@@ -9,19 +9,24 @@
 #include <stdbool.h>
 #include <time.h>
 #include <unistd.h>
+#include <string.h>
 #include <math.h>
 
 const char help[] =
-    "Usage: %s [-h] [-2] [-r rate]\n"
+    "Usage: %s [-h] [-2] [-c] [-r rate]\n"
     "  -h       print this help message and exit\n"
     "  -2       output two waves\n"
+    "  -c       randomly chunk the output\n"
     "  -r rate  sample rate in samples/s (default: 100)\n";
 
-const char optstring[] = "h2r:";
+const char optstring[] = "h2cr:";
 
 int main(int argc, char *argv[]) {
+    char buffer[1024];
+    size_t buffer_pos = 0;
     int opt;
     bool two_waves = false;
+    bool chunked = false;
     double rate = 100;
 
     // Parse the command line.
@@ -32,6 +37,9 @@ int main(int argc, char *argv[]) {
                 return EXIT_SUCCESS;
             case '2':
                 two_waves = true;
+                break;
+            case 'c':
+                chunked = true;
                 break;
             case 'r':
                 rate = atof(optarg);
@@ -49,10 +57,24 @@ int main(int argc, char *argv[]) {
     const useconds_t delay = 1e6 / rate;
 
     for (unsigned int n=0; ; n+=5) {
-        printf("%.1f\n", (sin(n*M_PI/180)*5)+5);
+        buffer_pos += sprintf(buffer + buffer_pos, "%.1f\n", (sin(n*M_PI/180)*5)+5);
         if(two_waves)
-            printf("%.1f\n", (cos(n*M_PI/180)*5)+5);
-        fflush(stdout);
+            buffer_pos += sprintf(buffer + buffer_pos, "%.1f\n", (cos(n*M_PI/180)*5)+5);
+        if (chunked) {
+            size_t send_pos = 0;
+            while (buffer_pos - send_pos >= 16) {
+                const size_t bytes_to_send = 1 + rand() % 16;  // 1..16
+                write(STDOUT_FILENO, buffer + send_pos, bytes_to_send);
+                usleep(50);  // let ttyplot read this before proceeding
+                send_pos += bytes_to_send;
+            }
+            if (send_pos > 0 && send_pos < buffer_pos)
+                memmove(buffer, buffer + send_pos, buffer_pos - send_pos);
+            buffer_pos -= send_pos;
+        } else {
+            write(STDOUT_FILENO, buffer, buffer_pos);
+            buffer_pos = 0;
+        }
         usleep(delay);
     }
 
--- a/stresstest.c
+++ b/stresstest.c
@@ -6,17 +6,44 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <stdbool.h>
 #include <time.h>
 #include <unistd.h>
 #include <math.h>
 
-int main(int argc, char *argv[]) {
-    (void) argv;
-    unsigned int n;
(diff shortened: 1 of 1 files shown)


torture.c
```diff
@@ -8,18 +8,72 @@
...
+
+    // Parse the command line.
+    while ((opt = getopt(argc, argv, optstring)) != -1) {
+        switch (opt) {
+            case 'h': printf(help, argv[0]); return EXIT_SUCCESS;
+            case '2': two_waves = 1; break;
+            case 'c': chunked = 1; break;
+            case 'g': add_garbage = 1; break;
+            case 'r': rate = atof(optarg); break;
+            default: fprintf(stderr, help, argv[0]); return EXIT_FAILURE;
```
hartwork, line 42: For readability please put newlines after colons and semicolons here. It's a bit too dense right now. Thanks!

torture.c
```diff
@@ -8,18 +8,72 @@
 #include <stdlib.h>
 #include <time.h>
 #include <unistd.h>
+#include <string.h>
 #include <math.h>
 
+char buffer[1024];
+size_t buffer_pos = 0;
```
hartwork, line 15: Can we pull these into `main` please? Did I miss a strong need for global variables here?

torture.c
```diff
@@ -8,18 +8,72 @@
...
+            buffer_pos += sprintf(buffer + buffer_pos, "garbage ");
+        if (two_waves) {
+            buffer_pos += sprintf(buffer + buffer_pos, "%.1f ", (cos(n*M_PI/180)*5)+5);
+            if (add_garbage && rand() <= RAND_MAX / 5)
+                buffer_pos += sprintf(buffer + buffer_pos, "garbage ");
+        }
+        if (chunked) {
+            size_t send_pos = 0;
+            while (buffer_pos - send_pos >= 16) {
+                int bytes_to_send = 1 + rand() % 16;  // 1..16
```
hartwork, line 64: I'd like to for `size_t` and `const` here.

torture.c
```diff
@@ -8,18 +8,72 @@
...
+            case 'r': rate = atof(optarg); break;
+            default: fprintf(stderr, help, argv[0]); return EXIT_FAILURE;
+        }
+    }
+    if (argc > optind) {
+        fprintf(stderr, help, argv[0]);
+        return EXIT_FAILURE;
+    }
+
+    useconds_t delay = 1e6 / rate;
```
hartwork, line 50: We have two places that sleep but only one variable `delay`. I'd like to vote for two with `const` or none.

torture.c
```diff
@@ -8,18 +8,72 @@
...
+
+    for (unsigned int n=0; ; n+=5) {
+        buffer_pos += sprintf(buffer + buffer_pos, "%.1f ", (sin(n*M_PI/180)*5)+5);
+        if (add_garbage && rand() <= RAND_MAX / 5)
```
hartwork, line 54: Since the new code calls `rand`, please also call `srand` with a value that can be overridden through a new CLI argument — e.g. `-s seed` — so that  deterministic operation remains possible when needed. Thank you!

hartwork (changes requested):
@edgar-bonet nice work, here's the small things I found:

hartwork:
@edgar-bonet I'm a bit worried that rebasing this into latest development with #118 merged now coulld kill my review comments above due to the rename. Maybe we can complete the pull request in isolation with the original filename and only rebase with a rename when it's done, to be safe; just an idea.

edgar-bonet:
All your comments make a lot of sense to me. Applied them all, thanks for the good suggestions. I also restored `\n` (instead of space) as a separator, as I had no good reason for changing it.

I also added a second commit to align the help message with your PR #120.

I am OK with completing this review round before rebasing on development.

torture.c
```diff
@@ -8,18 +8,92 @@
...
-        if(argc==2)
-            printf("%.1f\n", (cos(n*M_PI/180)*5)+5);
-        fflush(stdout);
-        usleep(10000);
+    char buffer[1024];
+    size_t buffer_pos = 0;
+    int opt;
+    int two_waves = 0;
+    int chunked = 0;
+    int add_garbage = 0;
```
hartwork, line 34: I noticed only now we have three more bools in disguise here. Can we turn these to stdbool also?

torture.c
```diff
@@ -8,18 +8,92 @@
...
+                chunked = 1;
+                break;
+            case 'g':
+                add_garbage = 1;
+                break;
+            case 'r':
+                rate = atof(optarg);
+                break;
+            case 's':
+                seed = atol(optarg);
```
hartwork, line 57: (Some compilers would complain here that we cast from `long` to `unsigned int` of potentially different size here. If we use `atoi` with "i" the size difference goes away or we could use an explicit cast to signal that we want to support bigger inputs even and feed them to the modulo wrap machine. Not too important, maybe :smiley: )

hartwork (changes requested):
> All your comments make a lot of sense to me. Applied them all, thanks for the good suggestions.

@edgar-bonet cool, thanks! 

> I also added a second commit to align the help message with your PR #120.

Very nice! :+1: 

Found two (likely last) minor thing more, let's see what you think.

hartwork:
> I integrated your last two comment that, while minor, were in my eyes relevant. Thanks for that! I also marked the PR as ready: I had labeled it “draft” only waiting for the rebase that would integrate the name change → stresstest.c. Should I do that rebase now?

@edgar-bonet sounds good, please go ahead, but the use of `bool` still seems to be missing while being resolve, did you forget to push something maybe?

hartwork:
@edgar-bonet nevermind bool, maybe me reload was broken, I see `bool` now, thanks! :+1

hartwork (approved):

edgar-bonet:
Rebased on `development`: now it's ready, the changed file is now `stresstest.c`.

hartwork (approved):

hartwork:
@edgar-bonet thanks, let's get this merged! :smiley:

PR #117, opened 2023-11-25
Introduce automated UI testing based on asciinema

hartwork:
This will prevent the next case of #102 right in the pull request introducing the regression.

This took quite a while to build, I ran into many interesting walls in the process and now finally have it working. Happy to answer questions about it.


.github/workflows/linux_and_macos.yml
```diff
@@ -99,6 +111,51 @@ jobs:
...
+        if: "${{ runner.os == 'macOS' }}"
+        run: |-
+          set -o pipefail
+
+          assert_images_equal_enough() {
+            local a="${1}"
+            local b="${2}"
+            local diff_output="${3}"
+            local dissimilarity="$(compare -metric DSSIM "${a}" "${b}" "${diff_output}" 2>&1)"
+            if ! python3 <<<"import sys; sys.exit(int(${dissimilarity} > 0.0022))"; then
```
edgar-bonet, line 137: Maybe `bc` could be a lightweight alternative to Python:

```shell
if [ $(bc <<< "$dissimilarity > 0.0022") = "1" ]; then ...
```

ttyplot.c
edgar-bonet, line 1: Instead of adding dedicated code to `ttyplot.c`, maybe you could wrap the call to `ttyplot` with [faketime][].

[faketime]: https://github.com/wolfcw/libfaketime

edgar-bonet (commented):
Automated visual inspection in CI? Wow! That's awesome! I find this really great, but at the same time I am a bit worried this could cause some friction for contributions that legitimately change the rendering. I understand in such a case one could push the change, wait for the CI to fail, download the generated artifacts, commit the actual images as the expected ones, and push again. It would be a bit cumbersome though.

Could one instead just run `record.sh` to get the expected images? I can't find the `agg` command on the Ubuntu repos, is there an alternative?

.github/workflows/linux_and_macos.yml
hartwork: @edgar-bonet I started out with `bc` in fact, but it turned out that ImageMagick's `compare` sometimes produces scientific output like `7.57917e-05` and `bc` fails to parse that (see below) while Python 3 has no problem with it:

```console
# bc <<<"7.57917e-05 < 123"
(standard_in) 1: syntax error

# python3 <<<"print(7.57917e-05 < 123)"
True
```

That's the only reason I went for Python there.

ttyplot.c
hartwork: @edgar-bonet I started out with `libfaketime` — that's what inspired the variable name — but `libfaketime` works with `LD_PRELOAD` and I (falsely?) assumed that `libfaketime` would not work on macOS then. I beleive I even seacrh homebrew for it, but it's there, not sure what happened, maybe that's why one should stop before 6am in the morning, this thing was quite addictive to get working. From a quick search again I may just be wrong. I'll definitely try again with `libfaketime` if the plan is to merge a version of this with `libfaketime`.

hartwork:
> Automated visual inspection in CI? Wow! That's awesome!

@edgar-bonet glad you like the idea!

> I find this really great, but at the same time I am a bit worried this could cause some friction for contributions that legitimately change the rendering. I understand in such a case one could push the change, wait for the CI to fail, download the generated artifacts, commit the actual images as the expected ones, and push again. It would be a bit cumbersome though. Could one instead just run `record.sh` to get the expected images?

Yes!

The fail-and-download approach would work, but I would personally see it as a fallback, as a second option, not the primary way, at least that was not intended. What I did locally to produce fresh images was basically this:

```bash
cd recordings/
./record.sh
for i in actual-*.png ; do cp ${i} ${i/actual/expected}; done
for i in expected-*.png ; do zopflipng -y ${i} ${i}; done  # https://github.com/google/zopfli
git add expected-*.png
```

I can make that a second script `sync.sh` or so if that changes something. Maybe without the `git add`to not interfere with the developers workflow. What do you think?

>  I can't find the `agg` command on the Ubuntu repos, is there an alternative?

`agg` is https://github.com/asciinema/agg and Gentoo doesn't have a package either. (It does have `x11-libs/agg` but that's somthing else). What I did and considered okay at least for myself was…

```
cargo install --git https://github.com/asciinema/agg
export PATH="${PATH}:$HOME/.cargo/bin"
```
…and then command `agg` is readily available.

ttyplot.c
hartwork: @edgar-bonet I have added a commit migrating to libfaketime now. It turns out that on macOS, GCC builds with ignore libfaketime while the Clang builds take it into account. The complexity is also significantly higher than before. I'm good with eitehr way forward, please have a look and let me know what you think.

hartwork: @edgar-bonet update on that: Turns out that both GCC and Clang builds ignore libfaketime some of the time, there is some weird flakiness going on here, maybe a suttle difference in runner images across CI runs. I'm reverting the the version before libfaketime for now, CI needs to be stable, the diff is still there for inspection or further experiments.

edgar-bonet: I still think `faketime`  would be a nicer solution, but if it is hard to set up reliably, we can move on and let that for another time.

CONTRIBUTING.md
```diff
@@ -0,0 +1,56 @@
...
+    cargo \
+    fonts-liberation
+$ cargo install --git https://github.com/asciinema/agg
+```
+
+…and put the install location of `agg` into `${PATH}`:
+
+```console
+$ export PATH="${PATH}:${HOME}/.cargo/bin"
+```
```
edgar-bonet, line 56: This seems easy enough. I tried, and it just worked. Thanks!

edgar-bonet (approved):
LGTM.

ttyplot.c
hartwork: @edgar-bonet thanks! I plan to investigate a bit more on the CI image side of things and may open a ticket to them about it. Gradual rollout is one thing and maybe better than breaking everyone at the same time but flipping images back and forth for the same repository every five minutes is not very nice to users of GitHub Actions. Let's see if I can confirm that impression.

PR #118, opened 2023-11-25
Rename tool "torture" to "stresstest" + build stresstest for "make" and "make all" also

hartwork:


edgar-bonet (approved):
Maybe it could make sense that `make` builds only `ttyplot` and `make all` builds both?

hartwork:
> Maybe it could make sense that `make` builds only `ttyplot` and `make all` builds both?

I think `all` is the de-facto default target and is hence closer to "default" semantically rather than "absolutely all".
Maybe in the interest of moving forward and with your approval given let's move forward with this and adjust once more if we learn that there are more reasons why we want something closer to the old behavior for reasons not know right now.

PR #119, opened 2023-11-26
Bump version to 1.6.0

hartwork:
With the slight differences in behavior and the number of improvements maybe "2.0.0" will be a better fit that "1.5.3", so I went for that. Just an idea, and also demos a pull request where UI tests are affected.


edgar-bonet (commented):
It makes sense to bump to at least 1.6. Maybe to 2.0 if introducing a behavior incompatible with previous versions, e.g. more tolerant parsing. In the end, I think the decision belongs to @tenox7.

tenox7:
I would prefer 1.6 rather than 2.0. I think I want to reserve 2.0 if we completely rewrite it in Go or Rust or something like that.

snap/snapcraft.yaml
```diff
@@ -1,5 +1,5 @@
 name: ttyplot
-version: 1.5.2
+version: 2.0.0
```
tenox7, line 2: I would prefer 1.6.x, thanks

hartwork:
> I would prefer 1.6 rather than 2.0. I think I want to reserve 2.0 if we completely rewrite it in Go or Rust or something like that.

@tenox7 alright, re-targeted to 1.6.0 now.

edgar-bonet (approved):

PR #121, opened 2023-11-26
Make CI reject spelling errors using codespell + fix three typos

hartwork:
This is a port of https://github.com/libexpat/libexpat/pull/739 to ttyplot. Should help keep our typo count down in the future.

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,28 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under Apache License Version 2.0
+
+name: Enforce codespell-clean spelling
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * 5'  # Every Friday at 3am
+  workflow_dispatch:
+
+# Minimum permissions for security
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Enforce codespell-clean spelling
+    runs-on: ubuntu-22.04
+    steps:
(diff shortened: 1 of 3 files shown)


.github/workflows/codespell.yml
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2023 Sebastian Pipping <sebastian@pipping.org>
+# Licensed under Apache License Version 2.0
+
+name: Enforce codespell-clean spelling
+
+on:
+  pull_request:
+  push:
+  schedule:
+    - cron: '0 3 * * 5'  # Every Friday at 3am
```
edgar-bonet, line 10: Why the schedules run? Is that in case the word list changes?

edgar-bonet (approved):

.github/workflows/codespell.yml
hartwork: @edgar-bonet the general idea that even if we try to pin things down as much as possible (which is the most secure and most robust approach meaning tas furthest away from flakey), some part of the CI ground below us is always moving and if we keep running CI once a week, we will never be more than two weeks behind noticing CI broke in the mean time (and hence get a chance to fix things in time and not when we want to release some hotfix a.s.a.p.), even if we do not do new pull requests for a few weeks more. I haven't checked in detail yet how much movement is inside the codespell action despite pinning. What do you think?

edgar-bonet: I see. makes sense.

hartwork: @edgar-bonet cool! :tada:

PR #123, opened 2023-11-26
CI: Start using AddressSanitizer (ASan) and UndefinedBehaviorSanitizer (UBSan)

hartwork:


edgar-bonet (approved):
Sanitizers are a thing that I should definitively use more often. Having them in CI is definitively great! :clap:

PR #127, opened 2023-11-27
recordings: Make record.sh ensure recent enough asciinema

hartwork:


edgar-bonet:
I don't really get why asciinema needs `--cols` and `--rows`. It normally works fine without that. I would argue that, providing terminal dimensions is really the job of the terminal emulator. In this case, the job of headless.py. Can't this script fix the terminal dimensions?

hartwork:
> I don't really get why asciinema needs `--cols` and `--rows`. It normally works fine without that. I would argue that, providing terminal dimensions is really the job of the terminal emulator. In this case, the job of headless.py. Can't this script fix the terminal dimensions?

@edgar-bonet I did not find a way that worked personally, I did try but rather ran into things like https://github.com/asciinema/agg/issues/62 than success. If you want to play with an alternative and make it work from, inside `headless.py` so that it works fine on both local Linux and in macOS CI that's welcome, but making sure that everyone runs the same version of asciinema will likely not only be more economic but also fix other issues from divergence in version that we may have in the future, also. What do you think?

edgar-bonet (approved):
OK. I may try later to get the terminal defined in the Python script (not much time now), as I don't like that much having software installations “local” to one project. In the mean time, this deserves merging IMO.

hartwork:
> OK. I may try later to get the terminal defined in the Python script (not much time now), as I don't like that much having software installations “local” to one project. In the mean time, this deserves merging IMO.

@edgar-bonet thanks! :+1:

PR #129, opened 2023-11-28
Fix over-eager buffer truncation

hartwork:
Regression from #98, may need coordination with #115.

Previously, with high input pressure where full `sizeof(input_buf) - 1` bytes could be read, all content would be dropped right after reading, and it could happen with every single read.

Example:
```console
# ./ttyplot < /dev/random
```

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -514,19 +514,18 @@ int main(int argc, char *argv[]) {
 
         // Read as much from stdin as we can (first read after select is non-blocking)
         if (stdin_can_be_read_without_blocking) {
+            // Last resort: truncate existing input if input line ever is
+            //              too long
+            if (input_len >= sizeof(input_buf) - 1 /* terminator */) {
+                input_len = 0;
+            }
+
             char * const read_target = input_buf + input_len;
             const size_t max_bytes_to_read = sizeof(input_buf) - (input_len + 1 /* terminator */);
             const ssize_t bytes_read_or_error = read(STDIN_FILENO, read_target, max_bytes_to_read);
 
             if (bytes_read_or_error > 0) {
                 input_len += bytes_read_or_error;
-
-                // Last resort: truncate existing input if input line ever is
-                //              too long
-                if (input_len >= sizeof(input_buf) - 1) {
-                    input_len = 0;
-                }
-
                 assert(input_len < sizeof(input_buf));
                 input_buf[input_len] = '\0';
             } else {


edgar-bonet:
Since this section is completely rewritten in #115. Fixing this now may not be very useful. And I guess it could just add merge conflicts.

hartwork:
> Since this section is completely rewritten in #115. Fixing this now may not be very useful. And I guess it could just add merge conflicts.

@edgar-bonet it would give me a chance to clean-up the mess I made earlier regarding truncation, to would separate bugfixes from refactoring and get the fix into `development` sooner. Toyota would like it :smiley: 

I was hoping that we could merge this (#129) and rebase #115 on top of it after. Conflict resolution is cheap here, it could even be automated, because/assuming that #115 does not have the same problem. Let me demo that with our very case on the shell:

```bash
# Start after the first big refactoring commit, that would show conflict
git checkout -b demo-free-conflict-resolution edgar-bonet-readonly/refactor-input~1

# Merge but keep our side because the refactored version does not have the same issue
EDITOR=true git merge --strategy=ours github-readwrite/fix-buffer-truncation 

# Extract a rebased patch
git diff github-readwrite/fix-buffer-truncation HEAD > rebased.patch

# Pretend to start all over
git reset --hard edgar-bonet-readonly/refactor-input~2

# Apply the rebase version, three commits:
git cherry-pick github-readwrite/fix-buffer-truncation  # 1
patch < rebased.patch
git add ttyplot.c
git commit -C edgar-bonet-readonly/refactor-input~1  # 2
git cherry-pick edgar-bonet-readonly/refactor-input  # 3

# Done
```

What do you think?

hartwork:
@edgar-bonet PS: For a cleaner version using `git commit-tree`…

```console
# First commit (i.e. start right at <fix-buffer-truncation>)
git checkout -b demo-free-conflict-resolution-2 hartwork-readwrite/fix-buffer-truncation

# Second commit, fake a merge with startegy "theirs" plus remaking original commit metadata
git commit-tree edgar-bonet-readonly/refactor-input~1^{tree} -p HEAD -m replaced-later
git merge --ff "$(git commit-tree edgar-bonet-readonly/refactor-input~1^{tree} -p HEAD -m replaced-later | tee /dev/stderr)"
EDITOR=true git commit --amend -C edgar-bonet-readonly/refactor-input~1

# Third commit, plain cherry-pick
git cherry-pick edgar-bonet-readonly/refactor-input
```

edgar-bonet (approved):

hartwork:
@edgar-bonet thank you, very kind! :pray:

PR #131, opened 2023-11-28
Be robust towards input containing null bytes (regression from #98)

hartwork:
Regression from #98, may need coordination with #115.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -418,6 +418,15 @@ bool handle_input_event(void)
         redraw_needed = true;  // redraw to display the error message
         return true;
     }
+
+    // The data we read could contain null bytes, so we replace those
+    // by one of the supported delimiters to not lose all input coming after.
+    for (size_t i = buffer_pos; i < buffer_pos + bytes_read; i++) {
+        if (buffer[i] == '\0') {
+            buffer[i] = ' ';
+        }
+    }
+
     buffer_pos += bytes_read;
 
     // Handle this new data.


edgar-bonet:
> may need coordination with #115

It indeed does. Your fix seems perfect to me. Should I add this to refactor-input? Or maybe let you rebase this PR once #115 gets merged?

hartwork:
> It indeed does. Your fix seems perfect to me. Should I add this to refactor-input? Or maybe let you rebase this PR once #115 gets merged?

@edgar-bonet I can rebase it onto #115 after merge but if acceptible I would prefer the other way around: merging #131 here first and rebasing #115 onto this after. It would reduce the number of things in flight, get the fix out earlier, and I'm imagining the conflict resolution to be easy here, if not I'd definitely volunteer to do the resolution. It would help us separate bugfixes from the refactoring, which I consider a good thing if feasible.

edgar-bonet:
> It would reduce the number of things in flight

That's exactly why I would like to merge #115 first. It has been “in flight” for quite a while now. But if you insist in getting this first, OK, go ahead.

hartwork:
@edgar-bonet you gave #129 to me where order was more important to me, I don't insist, let's do your #115 before my #131 here then, happy to rebase after, it's still a good deal for me :+1:

hartwork:
@edgar-bonet rebased and ready for review now

edgar-bonet (approved):
> The last part is a question of view, the former is a question of performance, at least in part. What do you think?

OK, it does make sense.

PR #132, opened 2023-11-28
Call `pselect` twice to not starve signal handling (alternative to #130)

hartwork:
Alternative to #130.

With both aggressive stdin input and signals coming in, `pselect` sometimes delays signal delivery for multiple seconds, to the point where even holding <kbd>Ctrl</kbd>+<kbd>C</kbd> down for seconds will not be successful in terminating the process. I have witnessed this symptom for a duration of 20+ seconds in practice. 

To reproduce:
```console
# ./ttyplot < /dev/zero
[hold Ctrl+C or press repeatedly]
[see it take multiple seconds for the process to terminate (if at all)]
```

Related:
https://stackoverflow.com/questions/62315082/pselect-on-linux-does-not-deliver-signals-if-events-are-pending

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -280,6 +280,32 @@ void finish(int signum) {
     sigint_pending = 1;
 }
 
+int pselect_without_signal_starvation(
+        int nfds,
+        fd_set * readfds,
+        fd_set * writefds,
+        fd_set * exceptfds,
+        const struct timespec * timeout,
+        const sigset_t * sigmask) {
+    // With high pressure on file descriptors (e.g. with "ttyplot < /dev/zero")
+    // a call to `pselect` could stall signal delivery for 20+ seconds on Linux.
+    // To avoid that situation, we first do a call to `pselect` that is dedicated
+    // to signal delivery and that only.
+    // (Related: https://stackoverflow.com/q/62315082)
+    const struct timespec zero_timeout = { .tv_sec = 0, .tv_nsec = 0 };
+
+    // First call, signal delivery only
+    const int select_ret = pselect(0, NULL, NULL, NULL, &zero_timeout, sigmask);
+
+    const bool signal_received = ((select_ret == -1) && (errno == EINTR));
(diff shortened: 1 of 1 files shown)


edgar-bonet:
This has a few issues:

* It would have to be rewritten after #115.

* It has a race condition for signals delivered between the two `pselect` calls.

* It makes the program repaint way more often than needed, even in the most common case of reasonable stdin pressure.

hartwork:
@edgar-bonet I understand that this is the one of the more controversial among the pull requests we had, and we don't have to rush this.

Regarding your points:

> * It would have to be rewritten after [Refactor the handling of input data #115](https://github.com/tenox7/ttyplot/pull/115).

Personally I would rather rebase #115 on top of this after #132 has been merged; the conflict resolution should be trivial, if it's not I would volunteer for the resolution.

> It has a race condition for signals delivered between the two `pselect` calls.

Yes, a race condition that will cause up to ~200ms delay in handling of Ctrl+C compared to 20+ seconds on current `development`.

> It makes the program repaint way more often than needed, even in the most common case of reasonable stdin pressure.

Yes, but so far I considered rendering at 5 fps acceptable, we can discus that or revert back to 500ms and be conscious about it.

To summarize, at the moment all these combined seem minor to me when compared to non-functioning Ctrl+C.

hartwork:
@edgar-bonet I have extracted a function `pselect_without_signal_starvation` now to restore the feel that there is one single event waiting location in the code. I hope you like it.

edgar-bonet (approved):
This still has a race condition. But the window of vulnerability is small, so this will do for now. Will probably implement a race-free alternative another day.

hartwork:
> This still has a race condition. But the window of vulnerability is small, so this will do for now. Will probably implement a race-free alternative another day.

@edgar-bonet perfect, thank you! :pray:

hartwork:
@edgar-bonet rebasing onto latest `development` for a green CI…

PR #136, opened 2023-11-30
Make functions and global variables static (e.g. to allow inlining)

hartwork:


edgar-bonet (approved):
Having all non-exported symbols `static` is definitively cleaner. Thanks for taking care of that!

PR #137, opened 2023-11-30
stresstest: Turn garbage mode from additive to corrupting

hartwork:
.. to help see that operation with `-2` does not mix up data from the two channels when some values are unparsable.

Try this before and after:
```console
# make
# ./stresstest -g -c -2 | ./ttyplot -2
```

### DIFF
--- a/stresstest.c
+++ b/stresstest.c
@@ -20,7 +20,7 @@ const char help[] =
     "  -h       print this help message and exit\n"
     "  -2       output two waves\n"
     "  -c       randomly chunk the output\n"
-    "  -g       occasionally output garbage\n"
+    "  -g       occasionally replace values by garbage\n"
     "  -r rate  sample rate in samples/s (default: 100)\n"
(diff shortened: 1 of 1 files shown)


edgar-bonet (approved):
Ideally, one should be able to do both. However, since ttyplot replaces invalid tokens with zeros, corrupting values is more relevant to the stress test. Although the current behavior of ttyplot may be open to discussion (maybe it could be better to ignore invalid tokens?), for now, this change is valuable.

hartwork:
@edgar-bonet ignoring invalid tokens would get us into a situation similar to what we had with `compare` of ImageMagick: some values do unexpected things — in that case it was switching to scientific notation for some values — and then channel two misses a value and suddenly you have channel one and two flipped. For something like network in and out that would be a disaster. Therefore I think we cannot skip values. Happy to discuss this more, merging for now due to the approval and in the interest of moving forward.

Issue #53, opened 2020-11-28
Support negative values


jason-green-io:
This would be great to plot temperature, or other values that typically sometimes go negative.

tenox7:
ack I will work on that

hartwork:
Related to #52

tenox7:
implemented in head/master

tenox7:
testing would be appreciated

tenox7:
👁️ ah I see it now thank you for pointing it out!

tenox7:
I think I have a working fix for this in https://github.com/tenox7/ttyplot/commit/f3bfe02843c9e64ef8398a8c3f6399ba867ce6e2

I also added printing 0 values when negative is present. I was debating this for long time but it appears to be standard. It looks little odd but I think it's valid. I'd appreciate your review and opinion.

<img width="777" height="415" alt="Image" src="https://github.com/user-attachments/assets/88bec5f5-b6b0-411c-8296-4625e3af900f" />

edgar-bonet:
@tenox7 wrote:
> I'd appreciate your review and opinion.

Nice job!

Now, negative values seem to work fine. I do also find that that the _z_ axis being “thick” looks odd. I would rather have it with zero width, but that may be just personal preference.

There may be a rounding issue going on here: the bottom of the plot seems to have been cropped. In the plot you showed, I noticed that the bottom of the sine wave is narrower than the top, and actually has the same with as the row just below the top. See the pale blue measurements:

<img width="777" height="415" alt="Image" src="https://github.com/user-attachments/assets/a50cb52e-fb43-4954-af7b-2417dc0ab326" />

I may just be nitpicking...

tenox7:
Yeah... I debated this and did some research. In most or all plot systems zero values, or zero crossing points do get plotted. The issue is that in graphical plots they are 1 pixel height or something negligible. 

Perhaps as an enhancement a zero line could be printed on the screen with ----------- characters. However technically it's better to see if a value had 0 value or not. It's possible that a value could have 0 value or not have zero value at some sample. Its better to see it than have it smeared with -----. I think.

Also for the bottom side being cropped I did notice that. I don't like it, but I would like to release with this and fix this rounding on bottom in a latter release.

edgar-bonet:
@tenox7 wrote:
> Also for the bottom side being cropped I did notice that. I don't like it, but I would like to release with this and fix this rounding on bottom in a latter release.

Makes a lot of sense to me.

PR #138, opened 2023-11-30
Deny negative values before proper support 

hartwork:
Regression from #115, related to #53

Try this before and after:
```console
# make
# ./ttyplot <<<'-1 -2 -3 -4'
```


edgar-bonet (commented):
Not sure what to make out of this. The current behavior is that ttyplot won't display negative values, but it does inform you that it has found values that are below the display range (`v` symbol). With this applied, ttyplot pretends that negative values are actually zero, except in rate mode: negative rates are still reported as being below the display range.

The current behavior seems more correct to me.

hartwork:
@edgar-bonet I agree that current `development` was on the right path but sneaking in 50% support for negative values does not seem right from a user point of view, e.g. with negative values down in the legend display where it's unclear where on the visual map they would be. Therefore the idea was to take a small step back, disable support explicitly, and then make it work properly for say 1.6.1 or 1.7.0. Right now it's a weird mix and none of the pull requests' titles say anything about it yet. Adding " + add some support for negative values" to the title of #115 for the change log would be the alternative, I guess. Would that work well for you? Other ideas?

hartwork:
@edgar-bonet made some major(?) edits to my reply above. Looking forward to your thoughts about it.

hartwork:
> Don't you think support for negative values would reach 1.6?

@edgar-bonet there is no reason why our ready work should be kept unreleased longer to get additional features ready, I like to separate things that do not _have_ to depend on each other. Personally, I will not work on support for negative numbers before 1.6.0 is out and it deserve proper thought, probably something like a vertically moving x-axis, or one snapping to a grid or or, it's not a five minutes job when done right, corner cases etc. I would personally like us to only fix any regressions towards 1.5.2 left more and maybe tiny trivial things and flush to users rather than making the cake later and bigger — it's big already — and still not have 1.6.0 by February or so. Many projects do that, e.g. KDbg 3.0.2 should be out already but isn't, exactly same scenario. The answer usually is to stay in releasable state all the time and to [release early, release often](https://en.wikipedia.org/wiki/Release_early,_release_often). How do you feel about it?

edgar-bonet:
> The answer usually is to stay in releasable state all the time

I concur, and I think we are keeping development in a releasable state. The reason this is not in master is because @tenox7 does not have right now the availability to keep up with this sprint of sorts. If I understood correctly, we are supposed to handle all this back to him only once the sprint is over.

> [...] still not have 1.6.0 by February or so.

I think it should definitely get out before Christmas.

> Personally, I will not work on support for negative numbers before 1.6.0 is out and it deserve proper thought

OK, I will give it a try.

> probably something like a vertically moving x-axis

No need. Some times one likes to have the _x_ axis drawn at _y_&nbsp;=&nbsp;0, but this is not a general rule, and certainly not a requirement.

PR #139, opened 2023-11-30
Avoid label "0.0 0.0 0.0 0.0" for y axis

hartwork:


edgar-bonet (approved):
Definitely looks cleaner.

PR #141, opened 2023-12-01
Document the key bindings

edgar-bonet:
Document them in the README, the man page, and the help message.

Note that I am not familiar with the troff language: I updated the man page just by copying the formatting from other sections. Hope it's correct. At least it renders OK.


ttyplot.1
```diff
@@ -83,6 +83,13 @@ Print the current version and exit.
 .It Fl h
 Print this help message and exit.
 .El
+.Sh KEY BINDINGS
+When reading data from a pipe, ttyplot accepts the following commands typed at the terminal:
```
hartwork, line 87: Could you make "ttyplot" bold as used in other places, e.g. in the authors section, using `.Nm` (probably)?

ttyplot.c
```diff
@@ -91,7 +91,9 @@ static void usage(void) {
             "  -t title of the plot\n"
             "  -u unit displayed beside vertical bar\n"
             "  -v print the current version and exit\n"
-            "  -h print this help message and exit\n\n");
+            "  -h print this help message and exit\n"
+            "\n"
+            "Interactive commands: q = quit, r = toggle rate mode\n");
```
hartwork, line 96: My vote for replacing "commands" by e.g. "hotkeys" and for making this last line three lines, for readability.

PS: The help output dump in the readme is no longer in sync with reality after this. Would be great to have in sync.

README.md
```diff
@@ -225,6 +225,21 @@ ttyplot also supports *counter* style metrics, calculating *rate* by measured ti
...
+
+
+when reading data from a pipe, ttyplot accepts the following commands typed at the terminal:
+
+```
+q   quit
+r   toggle "rate mode" on and off
+```
+
+these commands do not work if the standard input is a terminal: in this case quit with Ctrl-C.
```
hartwork, line 238: If this section is kept after syncing help output, please use `<kbd>...</kbd>` wraping for Ctrl and for C, each.

hartwork (changes requested):
@edgar-bonet thanks for this, that's a nice start. Here's what I found so far:

ttyplot.1
edgar-bonet: `.Nm` works. :smile:

ttyplot.c
edgar-bonet: OK, replaced with “hotkeys”. Note that the dump of the help output in the README is not a complete dump, as it lacks the line “Usage:”. Thus I used a dump of the second part of the help in the following section.

README.md
edgar-bonet: OK. Done.

ttyplot.c
hartwork: > OK, replaced with “hotkeys”.

@edgar-bonet thanks!

> Note that the dump of the help output in the README is not a complete dump, as it lacks the line “Usage:”. Thus I used a dump of the second part of the help in the following section.

@edgar-bonet I think midterm ideally it would be an identically copy. I have CI elsewhere to keep help output copies in sync…

- https://github.com/git-big-picture/git-big-picture/blob/main/.github/workflows/enforce-copies-of-help-output-in-sync.yml
- https://github.com/hartwork/git-delete-merged-branches/blob/master/.github/workflows/enforce-copies-of-help-output-in-sync.yml

…so something like that would be an option. It prevents cases like https://github.com/hartwork/resolve-march-native/pull/113. Midterm, just an idea, felt like mentioning it, a great case for something that CI can do in my opinion.

ttyplot.1
```diff
@@ -83,6 +83,15 @@ Print the current version and exit.
 .It Fl h
 Print this help message and exit.
 .El
+.Sh KEY BINDINGS
+When reading data from a pipe,
+.Nm
+accepts the following commands typed at the terminal:
```
hartwork, line 89: I find the word "command" really unfortunate for a hotkey. Can we replace it also here?

hartwork (changes requested):

README.md
hartwork: @edgar-bonet thanks! :+1:

README.md
```diff
@@ -225,6 +225,21 @@ ttyplot also supports *counter* style metrics, calculating *rate* by measured ti
...
+
+
+when reading data from a pipe, ttyplot accepts the following commands typed at the terminal:
+
+```
+  q  quit
+  r  toggle rate mode
+```
+
+these commands do not work if the standard input is a terminal: in this case quit with <kbd>Ctrl</kbd>-<kbd>C</kbd>.
```
hartwork, line 238: @edgar-bonet I think the dash should be a plus though?

ttyplot.1
edgar-bonet: I checked a few man pages of full-screen programs, and the word “commands” seems quite standard, much more than “hotkeys”:

* less(1) – **COMMANDS** 
* htop(1) – **INTERACTIVE COMMANDS** The following commands are supported [...]
* picocom(1) – **COMMANDS** Commands are given to picocom by [...]
* tmux(1) – **DEFAULT KEY BINDINGS** tmux may be controlled [...] followed by a command key.

ttyplot.c
```diff
@@ -91,7 +91,11 @@ static void usage(void) {
             "  -t title of the plot\n"
             "  -u unit displayed beside vertical bar\n"
             "  -v print the current version and exit\n"
-            "  -h print this help message and exit\n\n");
+            "  -h print this help message and exit\n"
+            "\n"
+            "Hotkeys:\n"
+            "  q  quit\n"
+            "  r  toggle rate mode\n");
```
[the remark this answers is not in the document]
edgar-bonet: It looks better indeed. Applied.

README.md
edgar-bonet: I've seen both, but the dash seems more common.

hartwork: @edgar-bonet okay

ttyplot.c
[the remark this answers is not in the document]
hartwork: @edgar-bonet cool, thanks! :pray:

hartwork (approved):
@edgar-bonet ready for prime time :+1:

PR #142, opened 2023-12-01
Catch signals on Linux with signalfd (follow-up to #98 and #132)

edgar-bonet:
As noticed by @hartwork, `pselect` is not a reliable way of catching signals on Linux: in case of high input pressure, signal delivery is starved. Pull request #132 partially fixed the issue by issuing two consecutive `pselect` calls, but this left a race window between these two calls.

This pull request fixes the issue without race conditions by using signalfd(), which makes signal delivery reliable. The call is only issued on Linux, as both the pselect() issue and signalfd() are Linux-specific.

This is one of two possible alternatives, the other one being the self-pipe trick (#143).

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -29,6 +29,9 @@
 #include <errno.h>
 #include <sys/types.h>
 #include <sys/stat.h>
+#ifdef __linux__
+#include <sys/signalfd.h>
+#endif
 #include <fcntl.h>
 
 #ifdef __OpenBSD__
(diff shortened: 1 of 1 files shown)


hartwork (commented):
@edgar-bonet on high level, I prefer #143 over this because it:
- has a lower chance of breaking support for BSD and/or Linux and/or other Unix through future changes because everyone is running the same code
- the control flow is simpler or rather there is only one flow rather than two in some sense

(One a low/detail level, two of my comments on #143 also apply here, but we can save those comments and changes if the plan is to get #143 ready to merge.)

Do you happen to have a preference for #142 here over the self-pipe trick, some benefit that I may be missing?

hartwork:
@edgar-bonet PS: nice demo, by the way :+1:

edgar-bonet:
@hartwork: This approach is indeed problematic from the point of view of testing. One trick I used is to `#undef __linux__` at the top, which lets me exercise the BSD code paths on Linux: as expected, it works, except for signals being missed under high stdin pressure.

I do also prefer the self-pipe approach: less code, and conceptually simpler as all events (signals, keystrokes and data) are caught in the exact same way: as “file descriptor ready” events. Closing this PR. I will work on #143, later, on the basis of your review.

PR #143, opened 2023-12-01
Catch signals using a self-pipe (follow-up to #98 and #132)

edgar-bonet:
As noticed by @hartwork, `pselect` is not a reliable way of catching signals on Linux: in case of high input pressure, signal delivery is starved. Pull request #132 partially fixed the issue by issuing two consecutive `pselect` calls, but this left a race window between these two calls.

This pull request fixes the issue without race conditions by using plain select() and a self-pipe, which makes signal delivery reliable.

This is one of two possible alternatives, the other one being `signalfd` (#142).

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -54,10 +54,7 @@
 #define T_LLCR ACS_LLCORNER
 #endif
 
-static sigset_t block_sigset;
-static sigset_t empty_sigset;
-static volatile sig_atomic_t sigint_pending = 0;
-static volatile sig_atomic_t sigwinch_pending = 0;
+static int signal_read_fd, signal_write_fd;
 static cchar_t plotchar, max_errchar, min_errchar;
 static struct timeval now;
 static double td;
@@ -281,40 +278,12 @@ static void paint_plot(void) {
     move(0,0);
 }
 
-static void resize(int signum) {
-    (void)signum;
-    sigwinch_pending = 1;
-}
-
-static void finish(int signum) {
-    (void)signum;
-    sigint_pending = 1;
-}
-
-static int pselect_without_signal_starvation(
-        int nfds,
-        fd_set * readfds,
-        fd_set * writefds,
-        fd_set * exceptfds,
-        const struct timespec * timeout,
-        const sigset_t * sigmask) {
-    // With high pressure on file descriptors (e.g. with "ttyplot < /dev/zero")
-    // a call to `pselect` could stall signal delivery for 20+ seconds on Linux.
-    // To avoid that situation, we first do a call to `pselect` that is dedicated
-    // to signal delivery and that only.
-    // (Related: https://stackoverflow.com/q/62315082)
-    const struct timespec zero_timeout = { .tv_sec = 0, .tv_nsec = 0 };
-
-    // First call, signal delivery only
-    const int select_ret = pselect(0, NULL, NULL, NULL, &zero_timeout, sigmask);
-
-    const bool signal_received = ((select_ret == -1) && (errno == EINTR));
-    if (signal_received) {
-        return select_ret;
-    }
-
-    // Second call
-    return pselect(nfds, readfds, writefds, exceptfds, timeout, sigmask);
+// Send signals through a pipe, in order to catch them without race conditions.
+// pselect() could be an alternative, but it is unreliable on Linux.
+// (Related: https://stackoverflow.com/q/62315082)
+static void signal_handler(int signum) {
+    const unsigned char signal_number = (unsigned char) signum;  // signum is either 2 (SIGINT) or 28 (SIGWINCH)
+    write(signal_write_fd, &signal_number, 1);
 }
 
 static void redraw_screen(const char * errstr) {
@@ -585,25 +554,28 @@ int main(int argc, char *argv[]) {
         cbreak();
     }
 
-    sigemptyset(&empty_sigset);
+    signal(SIGWINCH, signal_handler);
+    signal(SIGINT, signal_handler);
 
-    sigemptyset(&block_sigset);
-    sigaddset(&block_sigset, SIGINT);
-    sigaddset(&block_sigset, SIGWINCH);
-    sigprocmask(SIG_BLOCK, &block_sigset, NULL);
-
-    signal(SIGWINCH, resize);
-    signal(SIGINT, finish);
+    int signal_fds[2];
+    if (pipe(signal_fds) != 0) {
+        perror("pipe");
+        exit(1);
+    }
+    signal_read_fd = signal_fds[0];
+    signal_write_fd = signal_fds[1];
 
     while(1) {
         // Block until (a) we receive a signal or (b) stdin can be read without blocking
         // or (c) timeout expires, in order to reduce use of CPU and power while idle
         fd_set read_fds;
         FD_ZERO(&read_fds);
-        int select_nfds = 0;
+        FD_SET(signal_read_fd, &read_fds);
+        int select_nfds = signal_read_fd + 1;
         if (stdin_is_open) {
             FD_SET(STDIN_FILENO, &read_fds);
-            select_nfds = STDIN_FILENO + 1;
+            if (STDIN_FILENO >= select_nfds)
+                select_nfds = STDIN_FILENO + 1;
         }
         if (tty != -1) {
             FD_SET(tty, &read_fds);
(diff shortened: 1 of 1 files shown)


ttyplot.c
```diff
@@ -616,17 +584,21 @@ int main(int argc, char *argv[]) {
...
-            endwin();
-            initscr();
-            erase();
-            refresh();
-            gethw();
-            redraw_needed = true;
+        if (select_ret > 0 && FD_ISSET(signals[0], &read_fds)) {
+            int signum;
+            int count = read(signals[0], &signum, sizeof(signum));
+            if (count > 0) {
```
hartwork, line 590: Is there some guarantee here that read reads full `sizeof(int)` bytes every time? if yes, can we add an `assert`, if not, can we handle the case of `count > 0 && count < sizeof(signum)`?

ttyplot.c
```diff
@@ -54,10 +54,7 @@
 #define T_LLCR ACS_LLCORNER
 #endif
 
-static sigset_t block_sigset;
-static sigset_t empty_sigset;
-static volatile sig_atomic_t sigint_pending = 0;
-static volatile sig_atomic_t sigwinch_pending = 0;
+static int signals[2];  // pipe for sending signals to the main loop
```
hartwork, line 57: Just an idea: We could controduce two constants for read end (value 0) and write end (value 1) and use those in the code, to be further away from "why 0 here, what's the first one of the two again?".

ttyplot.c
```diff
@@ -616,17 +584,21 @@ int main(int argc, char *argv[]) {
...
-            sigwinch_pending = 0;
-            endwin();
-            initscr();
-            erase();
-            refresh();
-            gethw();
-            redraw_needed = true;
+        if (select_ret > 0 && FD_ISSET(signals[0], &read_fds)) {
+            int signum;
+            int count = read(signals[0], &signum, sizeof(signum));
```
hartwork, line 589: I would love `const` here :smile:

ttyplot.c
```diff
@@ -616,17 +584,21 @@ int main(int argc, char *argv[]) {
...
+            if (count > 0) {
+                if (signum == SIGINT) {
+                    break;
+                } else if (signum == SIGWINCH) {
```
hartwork, line 593: Just an idea: This `else` could be resolved for lower nesting level and arguably lower complexity.

hartwork (changes requested):
@edgar-bonet works well in practices, I will likely favor this over alternative #142 (..). Please see four tiny to small things below:

ttyplot.c
hartwork: @edgar-bonet PS: maybe `signal_fds` or `signalfds` would be a better name.

ttyplot.c
```diff
@@ -581,33 +549,33 @@ int main(int argc, char *argv[]) {
...
         }
         if (tty != -1) {
             FD_SET(tty, &read_fds);
             if (tty >= select_nfds)
                 select_nfds = tty + 1;
         }
-        struct timespec timeout = { .tv_sec = 0, .tv_nsec = 500e6 };  // 500 milliseconds for refreshing the clock
-        const int select_ret = pselect_without_signal_starvation(select_nfds, &read_fds, NULL, NULL, &timeout, &empty_sigset);
+        struct timeval timeout = { .tv_sec = 0, .tv_usec = 500e3 };  // 500 milliseconds for refreshing the clock
+        const int select_ret = select(select_nfds, &read_fds, NULL, NULL, &timeout);
```
hartwork, line 578: @edgar-bonet I only notice now that with this falling back to plain `select` rather than `pselect`, this would not fix the race but just move the race elsewhere? Am I missing something?

Also, if we do blocking reads and writes to the pipe, isn't that a problem? Do we need to put it in non-blocking mode and read as many signals as there are rather than just one?

ttyplot.c
edgar-bonet: My understanding is that `read` gives either the amount of data requested, or the amount available on the pipe, whichever is smaller. But now that you mention it, I am not too comfortable relying on that.

I changed the code to transmit a single byte, which is always atomic. We are only transmitting small numbers anyway: `SIGINT` is&nbsp;2, `SIGWINCH` is&nbsp;28, both on Linux and BSD. I do not know of any signal number larger than&nbsp;33.

ttyplot.c
edgar-bonet: Indeed, it's kind of a magic number. I made it clearer by using two separate variables: `signal_read_fd` and `signal_write_fd`.

ttyplot.c
edgar-bonet: Right. Done.

ttyplot.c
edgar-bonet: OK. Done.

ttyplot.c
edgar-bonet: I see no race here. Signals are never blocked, so the `write` to the pipe will always happen as soon as a signal arrives. And the signal will be handled on the next loop iteration.

Blocking reads should not block, as per the `select` promise. We may put all the file descriptors in non-blocking mode if we want to be extra defensive. Writes could only block if the pipe gets full, which can only happen if you send several thousand signals per second. You may be able to do that programmatically, but not by resizing the terminal or typing Ctrl-C.

ttyplot.c
hartwork: @edgar-bonet definitely more clear, thanks! :+1:

ttyplot.c
```diff
@@ -616,17 +588,21 @@ int main(int argc, char *argv[]) {
...
-            sigwinch_pending = 0;
-            endwin();
-            initscr();
-            erase();
-            refresh();
-            gethw();
-            redraw_needed = true;
+        if (select_ret > 0 && FD_ISSET(signal_read_fd, &read_fds)) {
+            unsigned char signal_number;
+            const int count = read(signal_read_fd, &signal_number, 1);
```
hartwork, line 593: Only notice now: Can we have `ssize_t` for a return type once more? it may differ from `int` in size.

ttyplot.c
```diff
@@ -277,40 +274,12 @@ static void paint_plot(void) {
...
-        return select_ret;
-    }
-
-    // Second call
-    return pselect(nfds, readfds, writefds, exceptfds, timeout, sigmask);
+// Send signals through a pipe, in order to catch them without race conditions.
+// pselect() could be an alternative, but it is unreliable on Linux.
+// (Related: https://stackoverflow.com/q/62315082)
+static void signal_handler(int signum) {
+    unsigned char signal_number = signum;
```
hartwork, line 281: @edgar-bonet this downcast is probably only safe because we know we only receive SIGWINCH and SIGINT here? I think it could use a comment, maybe even also a cast, I think MSVC would complain here.

ttyplot.c
hartwork: > I changed the code to transmit a single byte, which is always atomic.

@edgar-bonet very nice idea! :+1:

ttyplot.c
hartwork: > [..] Signals are never blocked [..]

@edgar-bonet while you say it like that, do we need to bring back signal blocking around interaction with ncurses a la pull request #130 then?

> [T]he `write` to the pipe will always happen as soon as a signal arrives. And the signal will be handled on the next loop iteration.
> 
> Blocking reads should not block, as per the `select` promise.

Understood, yes.

> We may put all the file descriptors in non-blocking mode if we want to be extra defensive. Writes could only block if the pipe gets full, which can only happen if you send several thousand signals per second. You may be able to do that programmatically, but not by resizing the terminal or typing Ctrl-C.

I mis-remembered then probably: I thought that writes to a pipe always block until there is a reader :thinking: `man 7 pipe` is with you.

edgar-bonet: > do we need to bring back signal blocking around interaction with ncurses a la pull request #130 then?

We don't. The signal handler does nothing but a `write`, which is async-signal-safe as per POSIX. The signal can really interrupt us any time without disrupting the program.

hartwork: @edgar-bonet isn't the important question whether the ncurses functions will stop their job half way when interrupted by a signal or not? From a quick look the man pages of `erase` and `refresh` say nothing about signals. I think that's what we'd need to know to be sure of no disruption. What do you think?

ttyplot.c
edgar-bonet: More generally, signal numbers are all small integers. I added a cast and a comment.

ttyplot.c
edgar-bonet: If an ncurses function is interrupted by a signal, it will stop halfway... then resume from where it stopped, and not even be aware that it had been interrupted. It is not very different from what happens to any running program during a hardware interrupt.

Most of these functions are not async-signal-safe, but this does _not_ mean it is unsafe to interrupt them: it means it is unsafe to call them from a signal handler.

That being said, it is conceivable that some ncurses function is buggy and mishandles a syscall failing with `EINTR`. In that case it would make sense to work around that bug by blocking signals around the call. However, I would avoid implementing workarounds for bugs we do not even know they exist.

hartwork: @edgar-bonet okay, let's note though that when we learn about that later, it will have been this move back to `select` from `pselect` that caused it and that — compared to 1.5.2 — now we have `select` without blocking signals while we had `select` *with* blocking signals in 1.5.2. There are multiple ways to see this and so going forward with #143 means that you are taking responsibility for dropping these blocks compared to 1.5.2; I cannot take the responsibility, because I cannot guarantee, that it will not cause new problems. Is that aware and okay with you? I'll do a last round of review then…

edgar-bonet: > we had `select` _with_ blocking signals in 1.5.2.

Let me take a look at 1.5.2 again... I see. That version called ncurses from the `SIGWINCH` signal handler, which is inherently unsafe and can lead, e.g., to memory corruption. This is reported in issue #62.

That version also blocked signals around calls to ncurses. This is a reasonable mitigation strategy: it prevents ncurses from being called on top of an interrupted ncurses function, and should thus prevent ncurses from corrupting its own state. The mitigation is not perfect though, and not guaranteed to always work. For example, it cannot prevent ncurses from calling `malloc` on top of an interrupted `malloc` call.

Now that we have a signal handler that is completely async-signal-safe, that mitigation has lost its purpose.

hartwork: @edgar-bonet I'll take that as "I'm sure enough that a block-do-unblock wrapper is not needed to take full responsibility for not re-adding some". Good.

PR #144, opened 2023-12-02
On demand timeout (follow-up to #98)

hartwork:
Follow-up to #98 in some sense

@edgar-bonet happy to adjust as needed or even close if this turns out to be the wrong way. Observed CPU load before and after was similar — jumping between 0.0%, 0.5% and 0.7%. Note the first commit. Looking forward to your review.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -455,6 +455,7 @@ static bool handle_input_event(void)
 int main(int argc, char *argv[]) {
     int i;
     bool stdin_is_open = true;
+    time_t prev_paint_at_seconds = 0;
     int cached_opterr;
     const char *optstring = "2rc:e:E:s:m:M:t:u:vh";
     int show_ver;
@@ -574,6 +575,7 @@ int main(int argc, char *argv[]) {
     gethw();
 
     redraw_screen(errstr);
+    prev_paint_at_seconds = now.tv_sec;
 
     // If stdin is redirected, open the terminal for reading user's keystrokes.
     int tty = -1;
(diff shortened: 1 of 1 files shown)


ttyplot.c
```diff
@@ -610,10 +612,24 @@ int main(int argc, char *argv[]) {
...
+        //    has advanced, only then redraw the clock.
         gettimeofday(&now, NULL);
+        int timeout_nanos = 0;
+        if (now.tv_sec == prev_paint_at_seconds) {
+            const suseconds_t microseconds_until_full_second = 1000 * 1000 - now.tv_usec;
+            timeout_nanos = 1000 * microseconds_until_full_second;
+        }
+        struct timespec timeout = {
+            .tv_sec = 0,
+            .tv_nsec = timeout_nanos
```
edgar-bonet, line 629: This could set `.tv_nsec` to 1e9, which should be avoided.

edgar-bonet (changes requested):
I would avoid calling `gettimeofday` more than once within a single loop iteration: this requires extra logic to check whether `tv_sec` changed between the calls, and brings no tangible benefit.

Keep in mind that everything in the program is non-blocking except `pslect`. Even on a slow machine, the whole loop should complete within a millisecond or so. In this case, the code gets conceptually simpler if we rely on the concept of “current iteration time”. This means neglecting the time spent on computing, and assuming the current time is a constant within each iteration. Then, the right place to call `gettimeofday` is right when `pselect` returns.

And yes, this could mean the clock update gets up to one millisecond late, but I wouldn't care too much about that.

ttyplot.c
hartwork: @edgar-bonet I don't see how, please help me see it.

hartwork:
> I would avoid calling `gettimeofday` more than once within a single loop iteration: this requires extra logic to check whether `tv_sec` changed between the calls, and brings no tangible benefit.
> 
> Keep in mind that everything in the program is non-blocking except `pslect`. Even on a slow machine, the whole loop should complete within a millisecond or so. In this case, the code gets conceptually simpler if we rely on the concept of “current iteration time”. This means neglecting the time spent on computing, and assuming the current time is a constant within each iteration. Then, the right place to call `gettimeofday` is right when `pselect` returns.
> 
> And yes, this could mean the clock update gets up to one millisecond late, but I wouldn't care too much about that.

@edgar-bonet I'm pushing a new commit demoing that single-`gettimeofday` appraoch now, but it doesn't feel right knowing that drawing and input processing takes more than zero time while it could easily be accounted for, it feels wrong, there is not other way to put it.

hartwork:
@edgar-bonet PS: I'm starting to think we should maybe go back to fixed 500ms, there was nothing truly wrong about it, and the complexity increase is missing payback in a way. How do you feel about that?

ttyplot.c
edgar-bonet: If `now.tv_usec` is zero, then `timeout_nanos` ends up being 1e9.

hartwork: @edgar-bonet my bad, classic case of "too obvious to see" :facepalm: Pushed.

hartwork:
Closing in favor of #145…

PR #145, opened 2023-12-02
Update the clock on the full seconds (follow-up to #98, alternative to #144)

edgar-bonet:
This is an alternative to #144, based on the idea of “current iteration time”. Since iterations are non-blocking and very fast (typically sub-millisecond), we can assume time is constant within each iteration. The global variable `now` holding this time is a loop invariant.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -610,7 +610,13 @@ int main(int argc, char *argv[]) {
             if (tty >= select_nfds)
                 select_nfds = tty + 1;
         }
-        struct timespec timeout = { .tv_sec = 0, .tv_nsec = 500e6 };  // 500 milliseconds for refreshing the clock
+        // Refresh the clock on the next full second.
+        const int nanoseconds_per_second = 1e9;
+        const int nanoseconds_remaining = nanoseconds_per_second - now.tv_usec * 1000;
+        struct timespec timeout = {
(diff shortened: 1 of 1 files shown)


ttyplot.c
```diff
@@ -610,7 +610,13 @@ int main(int argc, char *argv[]) {
             if (tty >= select_nfds)
                 select_nfds = tty + 1;
         }
-        struct timespec timeout = { .tv_sec = 0, .tv_nsec = 500e6 };  // 500 milliseconds for refreshing the clock
+        // Refresh the clock on the next full second.
+        static const int nanoseconds_per_second = 1e9;
```
hartwork, line 614: The `static` seems unnecessary here. I haven't checked whether it does damage, but it's not needed, my vote for dropping. Ideally, the compiler optimizes away all of it and then it doesn't use memory.

hartwork (changes requested):
@edgar-bonet let me think over this more, only one detail found so far:

ttyplot.c
edgar-bonet: Your are right, `static` does not bring us anything here. I'm with you: let's trust the compiler, at least for the very basic optimizations like this. Updated.

hartwork (approved):
@edgar-bonet tried it out, one of the simplest approaches we've had, alright, let's this merged :+1:

PR #146, opened 2023-12-03
Document the decision on <=1s timeout with `select` (follow-up to #145)

hartwork:
Regression from #145

@edgar-bonet this will not be popular, but please hear me out:

Our source the system clock updates at least once a second, and the current code sleeps up to one second. Because of these two combined, the user may see skips in seconds in the clock display, e.g. jumping from :03 straight to :05 or the clock taking longer than one second to update, i.e. running too slowly. Sleeping for a full second will never be right, and sleeping for 999e3 microseconds will not either, without really knowing how much time drawing takes in the worst case. I should have catched that during review, I'm sorry I didn't. Sleeping for 500ms not only was simple, it also didn't have this problem and it does give the CPU quite some air, it did not have a true problem that needed fixing. In the interest of both correctness and simplicity I would like to revert to where we were before, let's admit to ourselves that we took things too far. Please consider supporting this move. Thank you!

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -582,13 +582,29 @@ int main(int argc, char *argv[]) {
             if (tty >= select_nfds)
                 select_nfds = tty + 1;
         }
-        // Refresh the clock on the next full second.
+
+        // Refresh the clock on the next full second (plus a few milliseconds).
+        //
+        // We will sleep for a duration of up to a full second here knowing that:
+        // - we are technically putting two redraws apart by more than one second
+        // - that extra is only a few milliseconds (<25 in practice, often <1)
+        // - a few milliseconds is on the edge of what the human eye can notice
+        // - we can save CPU time (and potentially battery life) through giving
+        //   up on these milliseconds more in clock refresh delay accuracy.
+        //
+        // We had a constant timeout of 500 milliseconds before (which translates
+        // to twice the frequency of the maximum desired delay: redrawing at least
+        // once per second, the Nyquist frequency at work) but it ran the loop twice
+        // as much (including calling `select` twice as often) as the new approach.
+        // So we decided for lower CPU usage and a timeout of up to a full second.
+        //
         const int microseconds_per_second = 1e6;
         const int microseconds_remaining = microseconds_per_second - now.tv_usec;
         struct timeval timeout = {
             .tv_sec = microseconds_remaining / microseconds_per_second,
             .tv_usec = microseconds_remaining % microseconds_per_second
         };
+
         const int select_ret = select(select_nfds, &read_fds, NULL, NULL, &timeout);
 
         gettimeofday(&now, NULL);


hartwork:
@edgar-bonet I'm not sure I construct a case for the skipping — may not be possible —, but sleeping for 1 second is not right, we're going slower then the source then, and around to from select to select is not free. The user may not be able to see it, but it's wrong at the core, at math. Let us at least replace the `microseconds_remaining / microseconds_per_second` by 0 to be less wrong.

edgar-bonet:
Sorry, I still cannot see anything wrong. It looks to me like you are not comfortable with the idea of “current iteration time”, then you call it “wrong”.

I suggest you really try to build a scenario where the clock misbehaves. If you succeed, then we fix it. If you fail, the mere act of trying is likely to give you some insight, so the effort would not be wasted.

Now, one reason I dislike this PR is that the current code is all structured around the concept of _event loop_, and this PR goes against that concept. The core idea of the event loop is that we let the process sleep until there is an event that needs to be handled. “The clock display is out of date” is a valid event, which warrants waking up the process. We don't just wake up regularly “just in case”. In fact, the very first recommendation of the [Select tutorial][1] is:

> You should always try to use **select**() without a timeout. Your program should have nothing to do if there is no data available. Code that depends on timeouts is not usually portable and is difficult to debug.

Obviously, we cannot apply this recommendation verbatim – we depend on timeouts to detect the “clock display is out of date” event. But we can nonetheless follow its spirit and sleep until there is an actual event to process.

[1]: https://man7.org/linux/man-pages/man2/select_tut.2.html

hartwork:
> 👍
> 
> Note that, if this is still bothering you, the approach of #144 could be recycled. If we encapsulate the logic in a dedicated function, and we avoid carrying extra state, and we don't update `now` twice per iteration... then the main loop stays simple and sweet.

@edgar-bonet I see two key ideas combined here:
- Extracting a function
- Defensive protection against skips/jumps in display

Extracting the function will get some low-level bits out of the event loop, so that's nice.
With regard to protection against skips, I think the question is whether round trip…

```
select
gettimeofday
draw
select (again)
```

…can ever take a full second without getting to museum level hardware. If we can think of ways it can and have access to hardware that slow for testing, the defensive addition may help, if not it will just be zombie code stealing the CPU for everyone. Do you see a chance of that loop taking a full second? If yes, our whole "only a few milliseconds" assumption falls apart and then I would want us to say goodbye to sleeping full seconds altogether as a first priority. I think I'd vote for "yes" for extracting the function as a new pull request if you like and "no" for adding another check against lats painted second at the moment. What do you think?

> 
> ```c
> // Precondition: the current iteration time (now) has already been displayed.
> ```

That sentence seems to be mis-leading to me because "already" could mean anywhere and arbitrary long time ago.
We could maybe say that "`now` has been last updated between the last call to `select` and prior to the last paint (if any)" to be precise.

edgar-bonet:
> have access to hardware that slow for testing

I have an Acer Aspire One A110L (low-end netbook from 2008) and a first-generation Raspberry Pi from 2013. They could qualify as “slow”, at least relative to current hardware.

> Do you see a chance of that loop taking a full second?

The point of being non-blocking is to always keep this short. That being said, if the computer is _very_ heavily overloaded, I guess it can happen. But then the whole system would be slowed down to the point of being unusable, and a clock being laggy would be the least of the user's worries.

> I think I'd vote for "yes" for extracting the function as a new pull request if you like and "no" for adding another check against lats painted second at the moment. What do you think?

Not sure I understood...

hartwork:
> > I think I'd vote for "yes" for extracting the function as a new pull request if you like and "no" for adding another check against lats painted second at the moment. What do you think?
> 
> Not sure I understood...

@edgar-bonet I was trying to say that I would vote for extracting a function *without* adding an additional call to `gettimeofday`, a true and pure refactoring.

PR #149, opened 2023-12-04
Extract function `wait_for_events` (follow-up to #98 and #115)

hartwork:
Follow-up to #98 and #115

@edgar-bonet this is just an idea. I'm happy to rebase this onto #148 after its merge. Also, it doesn't block anything about your idea of extracting function `time_till_next_full_second` a la https://github.com/tenox7/ttyplot/pull/146#pullrequestreview-1762721004 before or after. Just an idea :beers:

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -430,6 +430,30 @@ static bool handle_input_event(void)
     return false;
 }
 
+// Refresh the clock on the next full second (plus a few milliseconds).
+//
+// We will sleep for a duration of up to a full second here knowing that:
+// - we are technically putting two redraws apart by more than one second
+// - that extra is only a few milliseconds (<25 in practice, often <1)
+// - a few milliseconds is on the edge of what the human eye can notice
+// - we can save CPU time (and potentially battery life) through giving
+//   up on these milliseconds more in clock refresh delay accuracy.
+//
+// We had a constant timeout of 500 milliseconds before (which translates
+// to twice the frequency of the maximum desired delay: redrawing at least
+// once per second, the Nyquist frequency at work) but it ran the loop twice
+// as much (including calling `select` twice as often) as the new approach.
+// So we decided for lower CPU usage and a timeout of up to a full second.
+//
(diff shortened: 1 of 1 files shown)


edgar-bonet (commented):
Well, why not. Could you please rebase on top of current `devel`. I wonder how it will come out. `EVENT_TIMEOUT` will probably have to be renamed (it's more like a “displayed clock is out of date” event) and given a non-zero value that could be or'ed with others. Also, this function may have to inherit the responsibility of updating the global `now`.

Regarding `time_till_next_full_second`, it could well be merged into `wait_for_events`.

edgar-bonet (approved):
It feels weird that `EVENT_TIMEOUT` is set but never used. But then, from the point of view of separation of concerns, it makes sense that `wait_for_events` doesn't care whether the reported events are ever used.

LGTM!

hartwork:
> It feels weird that `EVENT_TIMEOUT` is set but never used. But then, from the point of view of separation of concerns, it makes sense that `wait_for_events` doesn't care whether the reported events are ever used.

@edgar-bonet I feel the exact same way.


> LGTM!

Cool! Merging…

Is there anything left before a handover from us to Antoni?

edgar-bonet:
I was thinking of supporting negative values. However, I just got a flu, and my available energy is pretty low now. I think it is more reasonable to let that for another round. Let's get all this into Antoni's hands!

PS: Negative values are already somewhat supported, one just has to explicitly ask for them. For instance, `seq -10 0.2 10 | ttyplot -M -10` works just fine.

hartwork:
> I was thinking of supporting negative values. However, I just got a flu, and my available energy is pretty low now. I think it is more reasonable to let that for another round. Let's get all this into Antoni's hands!

@edgar-bonet my best wishes for a smooth recovery! Sounds fair, I'll reply to the related mail saying we'd like to hand over, you'll be in CC. It has less of a chance to go under in GitHub notification noise then, probably.

> PS: Negative values are already somewhat supported, one just has to explicitly ask for them. For instance, `seq -10 0.2 10 | ttyplot -M -10` works just fine.

Good to know!

PR #150, opened 2023-12-13
Keep the brace style consistent

edgar-bonet:
Function were indented in the 1TBS style, not K&R. Make new functions follow the same style.


hartwork (approved):
@edgar-bonet looks good! :+1:

PR #151, opened 2023-12-13
Remove function gethw()

edgar-bonet:
As we drop support for old systems, this function is not needed anymore. Call `getmaxyx()` directly instead, which is supposed to be always available.


hartwork (approved):
@edgar-bonet looks good! :+1:

PR #152, opened 2023-12-13
Make the CI enforce clang-format-clean code formatting

hartwork:


edgar-bonet (approved):
I don't know what to think about this one: clang-format is kind of a double-edged sword.

* On the one hand, it is a very effective way of guaranteeing style
  consistency. I do like this. On a personal note, I also do like the small changes made to the current style (spaces around operators, shorter lines…).

* On the other hand, I have witnessed this kind of CI action create
  friction for contribution. Especially from first-time contributors who almost inevitably oversee the .clang-format file and discover a failed CI.

I am approving, but this definitively needs to be pondered by @tenox7.

hartwork:
> * On the one hand, it is a very effective way of guaranteeing style consistency. I do like this. On a personal note, I also do like the small changes made to the current style (spaces around operators, shorter lines…).

@edgar-bonet cool!


> * On the other hand, I have witnessed this kind of CI action create friction for contribution. Especially from first-time contributors who almost inevitably oversee the .clang-format file and discover a failed CI.

When contributors hit this, the CI will produce a diff and it will clearly point to clang-format. I'm optimistic that C contributors will be able to figure out how to deal with it, it's literally just running `clang-format -i *.c` locally. We could treat is an an experiment and see how future contributors deal with it and collect data about it.

PR #154, opened 2023-12-19
Skip garbage tokens

edgar-bonet:
...instead of interpreting them as zeroes.


hartwork (approved):
@edgar-bonet looks good, thanks! :+1: 

Note to self: this reverted #137.

PR #157, opened 2023-12-25
 Fix crash with `-O1` on Ctrl+C + address `-O1` warnings + make CI compile with `-O1`

hartwork:
To reproduce:
    
1. Run `ttyplot` (as is)
2. Hit <kbd>Ctrl</kbd>+<kbd>C</kbd>
3. See message `*** bit out of range 0 - FD_SETSIZE on fd_set ***: terminate` and `Aborted`


.github/workflows/linux_and_macos.yml
edgar-bonet, line 1: Maybe add -O1 to the Makefile too?

stresstest.c
```diff
@@ -83,15 +83,15 @@ int main(int argc, char *argv[]) {
             size_t send_pos = 0;
             while (buffer_pos - send_pos >= 16) {
                 const size_t bytes_to_send = 1 + rand() % 16;  // 1..16
-                write(STDOUT_FILENO, buffer + send_pos, bytes_to_send);
+                (void)! write(STDOUT_FILENO, buffer + send_pos, bytes_to_send);
                 usleep(50);  // let ttyplot read this before proceeding
                 send_pos += bytes_to_send;
```
edgar-bonet, line 88: The compiler is telling us that we should check the return value of `write()`. Maybe we should listen to it instead of trying to silence it.

```suggestion
                const ssize_t bytes_sent = write(STDOUT_FILENO, buffer + send_pos, bytes_to_send);
                usleep(50);  // let ttyplot read this before proceeding
                if (bytes_sent > 0)
                    send_pos += bytes_sent;
```

stresstest.c
```diff
@@ -83,15 +83,15 @@ int main(int argc, char *argv[]) {
...
             }
             if (send_pos > 0 && send_pos < buffer_pos)
                 memmove(buffer, buffer + send_pos, buffer_pos - send_pos);
             buffer_pos -= send_pos;
         } else {
-            write(STDOUT_FILENO, buffer, buffer_pos);
+            (void)! write(STDOUT_FILENO, buffer, buffer_pos);
             buffer_pos = 0;
```
edgar-bonet, line 95:
```suggestion
            const ssize_t bytes_sent = write(STDOUT_FILENO, buffer, buffer_pos);
            if (bytes_sent > 0 && (size_t) bytes_sent < buffer_pos)
                memmove(buffer, buffer + bytes_sent, buffer_pos - bytes_sent);
            if (bytes_sent > 0)
                buffer_pos -= bytes_sent;
```

ttyplot.c
```diff
@@ -282,7 +282,11 @@ static void paint_plot(void) {
 // (Related: https://stackoverflow.com/q/62315082)
 static void signal_handler(int signum) {
     const unsigned char signal_number = (unsigned char) signum;  // signum is either 2 (SIGINT) or 28 (SIGWINCH)
-    write(signal_write_fd, &signal_number, 1);
+    while (true) {
+        const ssize_t write_res = write(signal_write_fd, &signal_number, 1);
+        if ((write_res == 1) || (errno != EINTR))
+            break;
+    }
```
edgar-bonet, line 289: Here, a `do...while` seems more intuitive to me than breaking out of an infinite loop.
```suggestion
    ssize_t write_res;
    do
        write_res = write(signal_write_fd, &signal_number, 1);
    while (write_res == -1 && errno == EINTR);
```

ttyplot.c
```diff
@@ -478,7 +482,7 @@ static int wait_for_events(int signal_read_fd, int tty, bool stdin_is_open, stru
             ret |= EVENT_SIGNAL_READABLE;
         }
 
-        if (FD_ISSET(tty, &read_fds)) {
+        if ((tty != -1) && FD_ISSET(tty, &read_fds)) {
```
edgar-bonet, line 485: Good catch!

edgar-bonet (approved):
Didn't notice the problem. Good think that you kept a sharp eye!

.github/workflows/linux_and_macos.yml
hartwork: @edgar-bonet that would actually cause problems to distros because they want full control over flags like that and then packaging would suddenly need ways around it, even patch it out if necessary. I have seen this with Gentoo multiple times. So only adding it to CI was intentional.

ttyplot.c
hartwork: @edgar-bonet it's a question of style and I don't like the need for moving the variable outside the loop personally, but since it's mostly a style preference, we can do your style, no problem. Let me try to make GitHub produce a commits from this and then turn them into squashes with you as the co-author. Let's see if that works out.

.github/workflows/linux_and_macos.yml
edgar-bonet: Oh, I see. Makes sense.

hartwork (commented):
@edgar-bonet I've added a few braces and changed whitespace in one spot but other than that this is your original suggestions squashed into my first commit and with co-author added. What do you think?

edgar-bonet (approved):

PR #160, opened 2024-01-02
Fixing some issues in Makefile that prevent a proper build

### LINKED ISSUE #159: ttyplot doesn't build (in general and on Void Linux) because of issues in the Makefile
I'm updating the Void Linux package and found several problems with the Makefile for which I provide a PR. Back quotes are not used for shell expansion in Make because a Makefile is not a shell script. The proper way is through `$(shell pkg-config ...)`. Also if CFLAGS are passed through `make CFLAGS=....` (like Void Linux does) the `+=` statements in Makefile have no effect. So I added `override` where it's needed.

MIvanchev:
Resolves #159

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,16 @@
 DESTDIR   ?=
 PREFIX    ?= /usr/local
 MANPREFIX ?= $(PREFIX)/man
+
+# Optional, thus no override.
+
 CFLAGS += -Wall -Wextra
-CFLAGS += `pkg-config --cflags ncursesw`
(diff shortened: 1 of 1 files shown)


hartwork (commented):
Let's maybe discuss this high level at https://github.com/tenox7/ttyplot/issues/159#issuecomment-1874196584 before diving into details for efficient use of everyone time.

hartwork:
Closing as "works as expected", more details at https://github.com/tenox7/ttyplot/issues/159#issuecomment-1874574254 .

PR #161, opened 2024-01-09
CI: Compile with `-std=c99` instead of `-std=gnu99`

hartwork:


edgar-bonet (approved):
It could make sense to put `-std=c99` in the Makefile rather than in the CI environment only.

hartwork:
> It could make sense to put `-std=c99` in the Makefile rather than in the CI environment only.

@edgar-bonet I'm unsure how these two approaches compare in effect. I can follow your suggestions if it is a suggestion. What difference in effect do you expect?

hartwork:
@edgar-bonet thanks for elaborating, I'm in the same "don't know for sure place". Should we merge as is and only move to the Makefile when the first issue comes in?

edgar-bonet:
> Should we merge as is and only move to the Makefile when the first issue comes in?

OK, makes sense to me.

PR #153, opened 2023-12-15
Add support for negative values

gsmecher:
Adds support for negative values. Should clear up #53 and #22.

This is a quick and possibly rough implementation - comments welcome.

PR #169, opened 2024-01-24
Add support for negative values

gsmecher:
This is an updated version of #153, and should clear up https://github.com/tenox7/ttyplot/issues/53 and https://github.com/tenox7/ttyplot/issues/22.

This is a quick and possibly rough implementation - comments welcome. The rebase looks clean enough, but I admit to rushing through it.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -74,7 +74,7 @@ static int signal_read_fd, signal_write_fd;
 static cchar_t plotchar, max_errchar, min_errchar;
 static struct timeval now;
 static double td;
-static double softmax = 0.0, hardmax = FLT_MAX, hardmin = 0.0;
+static double softmax = 0.0, hardmax = FLT_MAX, softmin = 0.0, hardmin = -FLT_MAX;
 static char title[256] = ".: ttyplot :.", unit[64] = {0}, ls[256] = {0};
 static double values1[1024] = {0}, values2[1024] = {0};
 static int width = 0, height = 0, n = -1, v = 0, c = 0, rate = 0, two = 0,
@@ -98,7 +98,8 @@ static void usage(void) {
         "  -e character to use for error line when value exceeds hardmax (default: e)\n"
         "  -E character to use for error symbol displayed when value is less than "
         "hardmin (default: v)\n"
-        "  -s initial scale of the plot (can go above if data input has larger value)\n"
+        "  -s initial maximum value (can go above if data input has larger value)\n"
+        "  -S initial minimum value (can go below if data input has smaller value)\n"
         "  -m maximum value, if exceeded draws error line (see -e), upper-limit of "
         "plot scale is fixed\n"
         "  -M minimum value, if entered less than this, draws error symbol (see -E), "
@@ -151,7 +152,7 @@ static void getminmax(int pw, double *values, double *min, double *max, double *
     int i = 0;
 
     *min = FLT_MAX;
-    *max = 0.0;
+    *max = -FLT_MAX;
 
     for (i = 0; i < pw && i < v; i++) {
         if (values[i] > *max)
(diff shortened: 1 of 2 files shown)


ttyplot.c
```diff
@@ -203,27 +201,43 @@ static void draw_line(int x, int ph, int l1, int l2, cchar_t *c1, cchar_t *c2,
...
+            l1 = 0;
+        else
+            l1 = lrint((v1[i] - min) / (max - min) * ph);
+
+        if (! v2 || isnan(v2[i]))
+            l2 = 0;
+        else if (v2[i] > hardmax)
+            l2 = ph;
+        else if (v2[i] < hardmin)
+            l2 = 0;
```
edgar-bonet, line 228:
```suggestion
        else if (v2[i] < hardmin)
            l2 = 1;
```
For symmetry with the case `v1[i] < hardmin`.

ttyplot.c
```diff
@@ -233,15 +247,19 @@ static void show_all_centered(const char *message) {
     mvaddnstr(y, x, message, width);
 }
 
-static int window_big_enough_to_draw(void) {
-    return (width >= 68) && (height >= 5);
+static int window_size_suitable(void) {
+    return (width >= 68) && (height >= 5) && (width < NSAMP);
 }
```
edgar-bonet, line 252: If the window size is not suitable, then an error message is displayed by `show_window_size_error()`: “Window too small...”. Now that “not suitable” can also mean “too big”, the error message should be adjusted.

ttyplot.c
```diff
@@ -203,27 +201,43 @@ static void draw_line(int x, int ph, int l1, int l2, cchar_t *c1, cchar_t *c2,
...
+
+    for (x = first_col; x < first_col + pw; x++, i = (i + 1) % pw) {
+        /* suppress drawing uninitialized entries */
+        if (! v1 || isnan(v1[i]))
+            continue;
+
+        if (v1[i] > hardmax)
+            l1 = ph;
+        else if (v1[i] < hardmin)
+            l1 = 0;
```
edgar-bonet, line 219:
```suggestion
        else if (v1[i] < hardmin)
            l1 = 1;
```
Ttyplot normally displays an “error symbol” when a value is less than `hardmin`. In order for this symbol to be visible, `l1` should be 1 here.

edgar-bonet (changes requested):
Thanks! This looks pretty nice to me. You even fixed a couple of errors in passing... Just a couple of minor things:

ttyplot.c
```diff
@@ -333,16 +325,18 @@ static void signal_handler(int signum) {
 }
 
 static void redraw_screen(const char *errstr) {
-    if (window_size_suitable()) {
+    if (width > NSAMP) {
```
edgar-bonet, line 328: This should be `plotwidth`:

* `width` is the total width of the terminal window
* `plotwidth`, which is set to `width - 4` by `paint_plot()`, is the width of the area used for plotting the data.

Within `plot_values()`, the data arrays are accessed through an index that wraps modulo `plotwidth`.

edgar-bonet (commented):
There are a couple things I should have noticed in my previous review, but they only strike me now... First is the confusion between `width` and `plotwidth` (see below). The second is that I do not understand the need for the check for `plotwidth > NSAMP`, as it seems redundant with this check near the top of `paint_plot()`:

```c
if (plotwidth >= (int)((sizeof(values1) / sizeof(double)) - 1))
    exit(0);
```

I did a test: starting from the head of this PR (163450c), I made `NSAMP` much smaller and removed the test `if (width > NSAMP)`. I compiled with `-fsanitize=address`, then ran `stresstest | ttyplot`. I enlarged the window while ttyplot was running and, when it got too large, ttyplot just exited (without properly resetting the terminal :unamused:). No memory errors. Repeated a few times, with repeatable results.

In the commit log of d1b0fa6, you wrote “Fix segfault with very wide windows”. Did you actually witness such segfault? If not, I suggest you drop this test for now.

There is still something that doesn't look quite right about the size checks, but it is the same here and on master, so I think it could be left for a subsequent pull request. While redrawing the screen, the normal sequence of events is:

1. `redraw_screen()` checks for the window not being too small (and also too large in this PR)
2. `paint_plot()` calls `getmaxyx()` to update our idea of the window size
3. `paint_plot()` checks for the window not being too large

Items 1 and 2 should be reversed in order for both checks to use the updated window size, as this is the one that will be used when actually plotting. There is also a race condition here if the window gets too small between 1 and 2, although the window of vulnerability seems tiny.

ttyplot.c
gsmecher: Hm, you are correct - this part of the PR looks misguided.

I do remember triggering a segfault, but in hindsight it's possible I triggered the exit(0) path and mistook it for something worse without following up.

This feels like a distraction - unless you have a preference, I think I'll back the overrange check out of the negative-value PR unless I can more carefully trigger a bug.

hartwork:
@gsmecher just a quick note that with the next push your CI here will be red due to meantime changes at https://apt.llvm.org/ *unless* you rebase onto latest `master` where pull request #170 with a fix has been merged. Thanks!

gsmecher:
> @gsmecher just a quick note that with the next push your CI here will be red due to meantime changes at https://apt.llvm.org/ _unless_ you rebase onto latest `master` where pull request #170 with a fix has been merged. Thanks!

Thanks - I have rebased and force-pushed, removing the bogus array size check at the same time.

edgar-bonet:
@gsmecher: There is something wrong with your last force-push. The previous one (77fb446) was, as expected, based on master. The last one (3d83c57) is based on master~ and carries copies of the two commits of #170, with different Committer name:

```text
* 3d83c57 (pull/169/head) Add support for negative values.
* 622f3a9 stresstest: add -n flag to remove positive bias (emit negative values)
* 6d24924 Don't use FLT_MIN - 0 is the correct value here.
* 1c579ef linux_and_macos.yml: Drop compilation with Clang 17
* bb6a97d linux_and_macos.yml: Get installation of Clang 18 on Linux back in sy>
* 5b64245 Move global variables that are only locally used into local scope.
| * 77fb446 Add support for negative values.
| * 0f1ae18 stresstest: add -n flag to remove positive bias (emit negative valu>
| * 216be4a Don't use FLT_MIN - 0 is the correct value here.
| * 0eb1356 Move global variables that are only locally used into local scope.
| * f98f2f2 (master) Merge pull request #170 from hartwork/fix-linux-ci
|/| 
| * 321ebd5 linux_and_macos.yml: Drop compilation with Clang 17
| * aef9d29 linux_and_macos.yml: Get installation of Clang 18 on Linux back in >
|/  
* 374bca3 readme update
```

Also, I had to unresolve a couple of review comments.

gsmecher:
My apologies. I am trying to keep this as a "side distraction" - which isn't fair if it pushes work onto you.

I have rebased onto f98f, and can see the correct commit history since then. I'll force-push that now, then look into the resolved/unresolved comments. Thanks for bearing with me.

edgar-bonet (approved):

tenox7 (approved):
Thank you!!! LGTM

tenox7:
Do you want me to merge?

tenox7 (approved):
LGTM

tenox7:
Thank you!

PR #181, opened 2024-05-13
Store the version string in a single place

edgar-bonet:
As shown by the recent version bumps (1.6.4 and 1.6.5), incrementing the version number is error prone. This is because this version number is stored in two different files (ttyplot.c and recordings/expected.txt), and these have to be kept manually in sync.

This pull request creates a single source of truth for the version string, which is the Makefile. The file expected.txt has been replaced by a template containing the string `@VERSION@`, which the CI workflow replaces by the actual version string. The script get\_back\_in\_sync.sh does the reverse replacement.


hartwork (commented):
@edgar-bonet I appreciate that you're trying to help with a true problem but in my eyes this needs a fix in process and no fix in Git: The CI protects us from having these things go out of sync unnoticed already and mass replacing the number is trivial and takes below one minute, which is good for something done as rarely in my book. So I'm having trouble approving something that fixes something that should be a non-problem already with a healthy workflow and it helps feeding `master` through direct pushes while it should only be fed through merged pull requests in a healthy green CI world. If the two of you want this, alright, but I consider the direct pushes to `master` the problem, personally.

MpcOS77 (approved):

edgar-bonet:
I wrote:
> As shown by the recent version bumps (1.6.4 and 1.6.5), incrementing the version number is error prone.

It appears this happened again on [the last][fail1] few [version bumps][fail2]. Whereas I agree with @hartwork that, in an ideal world, this should not be an issue, in practice it _is_ annoying.

Also, as the project owner, @tenox7 can choose tu push directly, without going through a pull request. We are not in a position to question this choice. It even appears that @tenox7 found that a good way to get rid of this annoyance was to simply [remove UI tests step from Linux/macOS workflow][commit].

As I believe these UI tests have some value, I have rebased this pull request on top of the current master branch, with an extra commit that reverts a079775c8a027223e7954c96b628aecd795bc7e9.

[fail1]: ../actions/runs/13951362293/job/39051473945#step:14:53
[fail2]: ../actions/runs/14280329908/job/40029108380#step:14:53
[commit]: ../commit/a079775c8a027223e7954c96b628aecd795bc7e9

tenox7 (approved):
LGTM

tenox7:
I like this. Approved, if you can resolve merge conflict. Also feel free to restore the version check with this.

edgar-bonet:
I just rebased the branch on top of master and force-pushed.

PR #187, opened 2024-07-10
Adds support for logarithmic y-axis scale

johanwiren:
Adds option "-l" to enable displaying the y-axis in log10 scale.


johanwiren:
Just noticed that negative values don’t render properly. I’ll open a new PR once I have fixed that.

PR #192, opened 2024-08-19
linux_and_macos.yml: Use gcc-12 instead of gcc-11

edgar-bonet:
The GitHub macOS 12 image has just dropped gcc-11: https://github.com/actions/runner-images/commit/4e6e715037c7824b785bfb8eb09f1eb59bdbf936

CC https://github.com/actions/runner-images/issues/10213

### DIFF
--- a/.github/workflows/linux_and_macos.yml
+++ b/.github/workflows/linux_and_macos.yml
@@ -37,7 +37,7 @@ jobs:
             clang_repo_suffix: null
             make: bmake
             runs-on: ubuntu-22.04
-          - cc: gcc-11
+          - cc: gcc-12
             clang_major_version: null
             clang_repo_suffix: null
             make: make
--- a/.github/workflows/linux_and_macos.yml
+++ b/.github/workflows/linux_and_macos.yml
@@ -23,44 +23,38 @@ jobs:
       matrix:
         include:
           - cc: gcc-14
-            cxx: g++-14
             clang_major_version: null
             clang_repo_suffix: null
             make: make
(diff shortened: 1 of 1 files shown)


.github/workflows/linux_and_macos.yml
```diff
@@ -40,7 +40,7 @@ jobs:
             clang_repo_suffix: null
             make: bmake
             runs-on: ubuntu-22.04
-          - cc: gcc-11
+          - cc: gcc-12
             cxx: g++-11
```
hartwork, line 44: Please also adjust this `g++-11` line. Or make a dedicated commit upfront deleting `cxx` altogether. What do you think?

hartwork (changes requested):

.github/workflows/linux_and_macos.yml
edgar-bonet: It looks to me like `cxx` is never used. I will remove it in a commit before this one (maybe after dinner).

hartwork: @edgar-bonet thanks! :+1:

hartwork (approved):
@edgar-bonet well done! :+1:

hartwork:
@edgar-bonet let's merge this first and then do "@dependabot rebase" comments in #190 and #191 so we get green CI everywhere — what do you think?

edgar-bonet:
> @edgar-bonet let's merge this first and then do `@dependabot rebase` comments in #190 and #191 so we get green CI everywhere — what do you think?

Seems good to me! 👍

PR #197, opened 2024-10-02
Fix compilation on FreeBSD

0mp:
On FreeBSD, we need to set _XOPEN_SOURCE_EXTENDED
(or __BSD_VISIBLE) to access everything we need.

A similar patch is already in use in the FreeBSD Ports:
https://github.com/freebsd/freebsd-ports/blob/main/graphics/ttyplot/files/patch-ttyplot.c.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -7,9 +7,9 @@
 // Apache License 2.0
 //
 
-// This is needed on macOS to get the ncurses widechar API, and pkg-config fails to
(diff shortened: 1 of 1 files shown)


0mp:
I've addressed the formatting issues reported by CI.

edgar-bonet (approved):
Looks good to me. Thanks for the PR!

Issue #193, opened 2024-08-19
The macOS 12 GitHub runner is getting deprecated


edgar-bonet:
The workflow `linux_and_macos.yml` performs three builds on macOS 12. GitHub announced today that [the macOS12 runner image is being deprecated][1]:

> We are beginning the deprecation process for the macOS 12 runner image, which allows us to balance our fleet capacity ahead of our upcoming [macOS 15 launch][2]. This image will be fully retired by the December 3rd, 2024. We recommend updating workflows to use `macos-14`, `macos-13`, or `macos-latest`.

[1]: https://github.blog/changelog/2024-08-19-notice-of-upcoming-deprecations-and-breaking-changes-in-github-actions-runners/
[2]: https://github.com/github/roadmap/issues/986

hartwork:
@edgar-bonet I guess we can (a) migrate to macos-13 now or (b) in December? How do you feel about it?

edgar-bonet:
@hartwork: I don't really care. I think both options are just fine.

hartwork:
@edgar-bonet I just tried macos-13 and macos-14 with GCC 12, 13, 14 and all I get is `ld: library not found for -lasan` or `ld: library 'asan' not found`. So I guess I'll vote for giving GitHub Actions time to fix their images with regard to AddressSanitizer until December.

tenox7:
deleted

hartwork:
It hurts to see c17d4da15842fe8713de860074007d2a2c0ec398 (and a079775c8a027223e7954c96b628aecd795bc7e9). GitHub Actions [has images for macOS 13 and 14](https://github.com/actions/runner-images) and giving up on macOS CI altogether (rather than on AddressSanatizer on macOS) was sure not what I was expecting (or suggesting). There is [ttyplot in Homebrew for macOS](https://formulae.brew.sh/formula/ttyplot) and without this CI you will not notice when you break things for macOS before but after releases. The loss of coverage of `bsdmake` is also collateral damage of this that seems unnecessary. I don't get it, I really don't 😞

tenox7:
It hurts for builds to fail every time because an UI test for a version that cannot be changed. I will accept a working PR for that.

PR #201, opened 2025-04-06
Restore test coverage on macOS

edgar-bonet:
The macOS tests were failing, presumably because of the deprecation of the GitHub macOS&nbsp;12 runners (c.f. #193). Then, they were disabled by commit c17d4da15842fe8713de860074007d2a2c0ec398.

This pull request restores those tests, while upgrading them to the supported runners: macOS 13, 14 and 15. It also bumps the tested versions of GCC and clang on those runners.


tenox7 (approved):
LGTM

tenox7:
merged thank you very much!

PR #204, opened 2025-09-22
Add _XOPEN_SOURCE_EXTENDED define to allow compiling on OpenBSD.

fcambus:
OpenBSD also needs the _XOPEN_SOURCE_EXTENDED define.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -6,9 +6,9 @@
 // Apache License 2.0
 //
 
-// This is needed on FreeBSD and macOS to get the ncurses widechar API,
-// and pkg-config fails to define it.
-#if defined(__APPLE__) || defined(__FreeBSD__)
+// This is needed on FreeBSD, OpenBSD, and macOS to get the ncurses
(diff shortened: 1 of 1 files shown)


edgar-bonet:
What about NetBSD and DragonFly BSD? Would it make sense to use a [test that matches all the BSDs][test]?

[test]: https://jdebp.uk/FGA/predefined-macros-platform.html#BSDs

fcambus:
I can't test on NetBSD or DragonFlyBSD, but I think testing for the `BSD` macro being defined isn't a good idea, in practice this idiom is rarely used in the wild.

I would suggest to keep this OpenBSD specific for now.

edgar-bonet (approved):
LGTM.

The CI failures are unrelated to this change. They are fixed by merging this into master.

fcambus:
Friendly ping on this one.

Since it has already been approved and the previous CI failures were noted as unrelated, would it be possible to merge it?

PR #214, opened 2026-03-08
Avoid division by zero on small windows

edgar-bonet:
If the window is too small when the program starts, ttyplot crashes on a division by zero (“Floating-point exception”). This is because `plotwidth` never gets set to a non-zero value, and `handle_value()` performs a modulo operation with `plotwidth` as the denominator.

This pull request fixes the crash by initializing `plotwidth` with the minimum value it can sensibly have.

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -53,6 +53,12 @@
 #define T_LLCR ACS_LLCORNER
 #endif
 
+// Window size
+#define WIDTH_MIN 68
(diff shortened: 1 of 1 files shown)


hartwork (approved):
@edgar-bonet nice one! I confirm crasher `COLUMNS=1 ./ttyplot <<<"1 2 3 4"` fixed.

tenox7:
thank you!

Issue #213, opened 2026-03-08
Window too small error in Termux on phone


Ytrog:
Is there a way to use it on smaller screens? I always get a "window too small" error on my phone.

Also, the 'q' button to quit it won't work when this error shows. I have to use `ctrl-c` *twice* to quit.

edgar-bonet:
@Ytrog: What is the size of your Termux terminal? You may get it with this command:

```shell
echo $COLUMNS x $LINES
```

Regarding the “window too small” error, ttyplot needs the terminal window to be at least 68×5. Maybe we could work with smaller windows by omitting the display of the date and time at the bottom.

Regarding the second issue (needing Ctrl-C), I cannot reproduce it on Ubuntu. Instead, I get a crash on “Floating-point exception”. I just filed pull request #214 in order to fix the crash. Could you please try it and tell us whether it makes the <kbd>q</kbd> responsive for you?

Ytrog:
The terminal size is 46 × 19

I'll look into the PR. I hope it is easy to install.

edgar-bonet:
@Ytrog wrote:
> I hope it is easy to install.

Assuming you cloned this repository with the remote name “origin” (which is the default), you may install the version in PR #214 like this:

```shell
git fetch origin refs/pull/214/head:pull-214
git checkout pull-214
make
sudo make install
```

Ytrog:
Thanks. That seems easy enough 😊👍

edgar-bonet:
@Ytrog: Hi! Did you have a chance of testing PR #214? Could you try #215, which specifically addresses this issue?

Ytrog:
Not yet. I was a bit busy. My apologies 🫣

Ytrog:
@edgar-bonet #214 I tested and the ctrl-c issue is resolved. Window too small not yet, but I'm going to test that now on #215.

Ytrog:
It works now if I apply #215

![Screenshot_20260319_105144_Termux.jpg](https://github.com/user-attachments/assets/48bf1cda-7f0a-46ad-b0ac-dc0ec5793e84)

Notice: I don't have `sudo` on Termux (non-rooted phone), so I tested it in the local folder instead.

PR #215, opened 2026-03-15
Work on smaller windows with no clock display

edgar-bonet:
This pull request addresses issue #213. When the terminal window width is less than 68 character cells, ttyplot refuses to plot anything. This limit comes from the space needed by the stuff displayed at the bottom which, at the minimal window width, make look like this:

```text
     │ last=8.5 min=5.0 max=8.5 avg=6.8    Sun Mar 15 15:45:37 2026
                            https://github.com/tenox7/ttyplot 1.7.4

```

This pull request alleviates the issue by removing the clock when the window is too narrow. The minimal window width is then shrunk by 24 cells, which is the width of the clock display.

There are a couple of caveats:

 * If the plotted numbers are very large, the last/…/avg display may overflow its allotted space.

 * In two-curve mode, the last/…/avg display for the second curve can collide with the ttyplot URL.

These issues are not specific to this PR though: they exist in the current master branch. For example, with a window width of 68:

**large numbers**:

```text
     │ last=18024.4 min=-31410.9 max=31420.9 avg=-3241.2  7:51 2026
                            https://github.com/tenox7/ttyplot 1.7.4

```

**two-curves**:

```text
     │ last=3.7 min=3.7 max=10.0 avg=7.8   Sun Mar 15 16:07:55 2026
       last=0.2 min=0.0 max=10.0 avg=4.6    om/tenox7/ttyplot 1.7.4
```

Possible improvements:

 * also remove the URL, or replace it with “ttyplot $VERSION”, if needed

 * print numbers in a format with a guaranteed maximum width (e.g. `%.3g`).

### DIFF
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -537,13 +537,14 @@ static void paint_plot(void) {
     if (width >= WIDTH_CLOCK_MIN) {
         const char *clock_display;
         if (fake_clock) {
-            clock_display = "Thu Jan  1 00:00:00 1970 ";
+            clock_display = "Thu Jan  1 00:00:00 1970";
         } else {
             lt = localtime(&now.tv_sec);
             asctime_r(lt, ls);
+            ls[strlen(ls) - 1] = '\0';  // drop trailing newline, see asctime_r(3)
             clock_display = ls;
         }
-        mvaddstr(height - 2, width - strlen(clock_display), clock_display);
+        mvaddstr(height - 2, width - strlen(clock_display) - 1, clock_display);
     }
 
     if (colors[TEXT_COLOR] != -1)
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -54,7 +54,8 @@
 #endif
 
 // Window size
-#define WIDTH_MIN 68
+#define WIDTH_MIN 44
+#define WIDTH_CLOCK_MIN 68
 #define WIDTH_MARGIN 4
 #define HEIGHT_MIN 5
 #define HEIGHT_MARGIN 4
@@ -532,15 +533,17 @@ static void paint_plot(void) {
 
     mvaddstr(height - 1, width - strlen(verstring) - 1, verstring);
 
-    const char *clock_display;
-    if (fake_clock) {
-        clock_display = "Thu Jan  1 00:00:00 1970 ";
-    } else {
-        lt = localtime(&now.tv_sec);
-        asctime_r(lt, ls);
-        clock_display = ls;
+    if (width >= WIDTH_CLOCK_MIN) {
+        const char *clock_display;
+        if (fake_clock) {
+            clock_display = "Thu Jan  1 00:00:00 1970 ";
+        } else {
+            lt = localtime(&now.tv_sec);
+            asctime_r(lt, ls);
+            clock_display = ls;
+        }
+        mvaddstr(height - 2, width - strlen(clock_display), clock_display);
     }
-    mvaddstr(height - 2, width - strlen(clock_display), clock_display);
 
     if (colors[TEXT_COLOR] != -1)
         attroff(COLOR_PAIR(TEXT_COLOR + 1));
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -54,8 +54,9 @@
 #endif
 
 // Window size
+#define WIDTH_CLOCK 24  // strlen("Thu Jan  1 00:00:00 1970")
 #define WIDTH_MIN 44
-#define WIDTH_CLOCK_MIN 68
+#define WIDTH_CLOCK_MIN (WIDTH_MIN + WIDTH_CLOCK)
 #define WIDTH_MARGIN 4
 #define HEIGHT_MIN 5
 #define HEIGHT_MARGIN 4


edgar-bonet:
@Ytrog just tested this PR on his phone running Termux. It does solve #213.

ttyplot.c
```diff
@@ -54,7 +54,8 @@
 #endif
 
 // Window size
-#define WIDTH_MIN 68
+#define WIDTH_MIN 44
+#define WIDTH_CLOCK_MIN 68
```
hartwork: I wonder if we can make these numbers explain themselves more, maybe?:

```suggestion
#define WIDTH_CLOCK 24  // strlen("Thu Jan  1 00:00:00 1970")
#define WIDTH_MIN 44
#define WIDTH_CLOCK_MIN (WIDTH_MIN + WIDTH_CLOCK)
```

ttyplot.c
```diff
@@ -532,15 +533,17 @@ static void paint_plot(void) {
...
-    if (fake_clock) {
-        clock_display = "Thu Jan  1 00:00:00 1970 ";
-    } else {
-        lt = localtime(&now.tv_sec);
-        asctime_r(lt, ls);
-        clock_display = ls;
+    if (width >= WIDTH_CLOCK_MIN) {
+        const char *clock_display;
+        if (fake_clock) {
+            clock_display = "Thu Jan  1 00:00:00 1970 ";
```
hartwork: While not 100% related, is the trailing space a bug here or am I missing something about `localtime`?:

```suggestion
            clock_display = "Thu Jan  1 00:00:00 1970";
```

hartwork (commented):
@edgar-bonet interesting idea!

ttyplot.c
edgar-bonet: Indeed, it is way clearer this way. I take the suggestion!

ttyplot.c
edgar-bonet: Oh, wait, good question. Let me see... This comes from commit 491c7768ff. I just did a test, and it turns out `asctime_r()` ends the string with `'\n'`. This behavior is actually specified in the manual page of `asctime()`, both in the [Linux library function manual][man3], and in the [POSIX Programmer's Manual][man3p], so I guess we can reliably count on it. The trailing space is then needed in order to get the correct value for `strlen(clock_display)` which is used below. Now, wondering what we could do about this... we may:

 (a). replace the trailing space by a `'\n'`, in order to be consistent with what `asctime_r()` does

 (b). let `gmtime()` and `asctime_r()` build the 1970 timestampt out of a zero `time_t`, in order to more closely mimic the `!fake_clock` case

 (c). remove the trailing space, then remove also the trailing `'\n'` in the output of `asctime_r()` in the `!fake_clock` case, and adjust the second argument of `mvaddstr()` accordingly

 (d). leave it as is, at least for this PR, and maybe address it in another PR

What do you think?

[man3]: https://man7.org/linux/man-pages/man3/asctime.3.html
[man3p]: https://man7.org/linux/man-pages/man3/asctime.3p.html

hartwork: @edgar-bonet interesting, thanks for elaborating on your findings!
From laying with it, I can confirm that `mvaddstr` treats the `\n` as a line feed and that the line feed going into `strlen` "happens" to match the math for "a border of width one" on the right. So in one case the border is respected by "drawing" a space over it and in the other by advancing the line at that moment.

I'd like to add one more option "(e) Using a space consistently":

```diff
diff --git a/ttyplot.c b/ttyplot.c
index 9c37605..1079e9c 100644
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -538,6 +538,7 @@ static void paint_plot(void) {
     } else {
         lt = localtime(&now.tv_sec);
         asctime_r(lt, ls);
+        ls[strlen(ls) - 1] = ' ';  // was a newline before, see asctime_r(3)
         clock_display = ls;
     }
     mvaddstr(height - 2, width - strlen(clock_display), clock_display);
```

For your (c) I believe the diff could be:

```diff
diff --git a/ttyplot.c b/ttyplot.c
index 9c37605..61c2cdb 100644
--- a/ttyplot.c
+++ b/ttyplot.c
@@ -534,13 +534,14 @@ static void paint_plot(void) {
 
     const char *clock_display;
     if (fake_clock) {
-        clock_display = "Thu Jan  1 00:00:00 1970 ";
+        clock_display = "Thu Jan  1 00:00:00 1970";
     } else {
         lt = localtime(&now.tv_sec);
         asctime_r(lt, ls);
+        ls[strlen(ls) - 1] = '\0';  // drop trailing newline, see asctime_r(3)
         clock_display = ls;
     }
-    mvaddstr(height - 2, width - strlen(clock_display), clock_display);
+    mvaddstr(height - 2, width - strlen(clock_display) - 1, clock_display);
 
     if (colors[TEXT_COLOR] != -1)
         attroff(COLOR_PAIR(TEXT_COLOR + 1));
```

I think (b) and its complexity is my least favorite, and I'd be happy with any of the other approaches :+1: 

PS: Maybe this `- 1` here means that `#define WIDTH_CLOCK_MIN (WIDTH_MIN + WIDTH_CLOCK + 1)` with `+ 1` would be needed for display of the clock?

edgar-bonet: > Maybe this `- 1` here means that `#define WIDTH_CLOCK_MIN (WIDTH_MIN + WIDTH_CLOCK + 1)` with `+ 1` would be needed for display of the clock?

I do not think so. My intent was to remove the display of the clock, but keep the one-column margin.

I will go for (c), and I find it cleaner to not print any type of space character after the clock.

Ytrog: Thank you all for the eﬀorts 😃

I hope these changes will also be distributed with Termux🤞

hartwork: > I do not think so. My intent was to remove the display of the clock, but keep the one-column margin.

@edgar-bonet I think you're essentially saying that `WIDTH_MIN` already includes that right margin of width one — I see!

hartwork (approved):

